PR #57, opened 2020-04-16
Man pages

alvarosanchez:


jameskleeh:
Should this be in the root project instead of core?

alvarosanchez:
I put this along with the Picocli dependencies and commands. However I see now that over the last days the repo has changed substantially. I'll resolve conflicts first and see where this would fit better

PR #98, opened 2020-04-22
refactor: Use name qualifier for AvailableFeatures

sdelamo:
Also, Ability to pass to ContextFixture:getFeatures and application type other than ApplicationType.DEFAULT

It deletes `ApplicationType::availableFeaturesClass`. 

In addition to the changes in this PR: I would like to delete classes: 

- `DefaultAvailableFeatures`
- `CliAvailableFeatures`
- `GrpcAvailableFeatures`
- `MessagingAvailableFeatures`
- `GrpcAvailableFeatures`

and just have: 

```java
package io.micronaut.starter.feature;

import io.micronaut.context.annotation.Factory;
import io.micronaut.core.order.OrderUtil;
import io.micronaut.starter.application.ApplicationType;

import javax.inject.Named;
import javax.inject.Singleton;
import java.util.Collection;
import java.util.List;
import java.util.stream.Collectors;

@Factory
public class AvailableFeaturesFactory {

    private List<Feature> features;

    public AvailableFeaturesFactory(Collection<Feature> features) {
        this.features = features
                .stream()
                .sorted(OrderUtil.COMPARATOR)
                .collect(Collectors.toList());
    }

    @Named("default")
    @Singleton
    public AvailableFeatures defaultAvailableFeatures() {
        return new AvailableFeaturesImpl(features, ApplicationType.DEFAULT);
    }

    @Named("cli")
    @Singleton
    public AvailableFeatures cliAvailableFeatures() {
        return new AvailableFeaturesImpl(features, ApplicationType.CLI);
    }

    @Named("function")
    @Singleton
    public AvailableFeatures functionAvailableFeatures() {
        return new AvailableFeaturesImpl(features, ApplicationType.FUNCTION);
    }

    @Named("grpc")
    @Singleton
    public AvailableFeatures grpcAvailableFeatures() {
        return new AvailableFeaturesImpl(features, ApplicationType.GRPC);
    }

    @Named("messaging")
    @Singleton
    public AvailableFeatures messagingAvailableFeatures() {
        return new AvailableFeaturesImpl(features, ApplicationType.MESSAGING);
    }
}
```

However, I think we need a concrete class for the picocli completion. 

That it is to say, we currently do: 

```java
   @CommandLine.Option(... completionCandidates = CliAvailableFeatures.class)
    @ReflectiveAccess
    List<String> features = new ArrayList<>();
```

and I don't think we can do:

```java
   @CommandLine.Option( .... completionCandidates = @Named("cli") AvailableFeatures.class)
    @ReflectiveAccess
    List<String> features = new ArrayList<>();
```


graemerocher:
Failures and conflicts

sdelamo:
@jameskleeh I've seen that you reverted the rename I did. 

Don't you think a method which returns a stream of visible features is better named 

` public Stream<Feature> getVisibleFeaturesStream()`

than:

` public Stream<Feature> getFeatures()`

jameskleeh:
@sdelamo Yes. Specifying the return type as part of the method name is redundant. The person calling the method can see the return type.

sdelamo:
what about the `visible` part of the method name?

jameskleeh:
Visibility is more of a detail of the features class. Other than internal usages, there should be no reason for anyone to call `getAllFeatures()`. So its left out of the method name because its an implementation detail.

Issue #111, opened 2020-04-22
AWS Custom runtime Feature


sdelamo:
See [AWS Lambda Custom runtime](https://micronaut-projects.github.io/micronaut-aws/snapshot/guide/index.html#lambdaCustomRuntimes)

PR #112, opened 2020-04-22
feat: Adds feature aws-lambda-custom-runtime

sdelamo:
See: https://github.com/micronaut-projects/micronaut-starter/issues/111

I flag it as a draft PR since it has only the dependencies. 

Moreover, test will fail because master is broken.


starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
```diff
@@ -331,6 +331,9 @@ dependencies {
 @if (features.contains("alexa-function")) {
     @dependency.template("io.micronaut.aws", "micronaut-function-aws-alexa", "implementation", null)
 }
+@if (features.contains("aws-lambda-custom-runtime")) {
+    implementation("io.micronaut.aws:micronaut-function-aws-custom-runtime")
```
jameskleeh, line 335: needs modified to use @dependency.template

PR #118, opened 2020-04-23
feat: Adds apt-idea

sdelamo:
see: https://plugins.gradle.org/plugin/net.ltgt.apt-idea


graemerocher:
the plugin is deprecated and not need to configure IDEA correct

PR #120, opened 2020-04-23
feat: apply gradle java-library plugin when not app

sdelamo:
When we don't apply the `application` Gradle plugin, I think when should apply the `java-library`


graemerocher:
I don't think we can assume that can we? Functions, messaging, CLI apps are not apps either

sdelamo:
But for functions. I want `java-library` applied. Don't I want it for other apps?. I mean we need to tell that the build contains java sources.

sdelamo:
For Kotlin, when we don't apply `application` gradle plugin we should apply the java plugin `id "java"` for groovy and java we can use `id "java-library"`

graemerocher:
A function is not a library, why would you want it applied?

graemerocher:
Unless you mean the "java" plugin which is not the same as "java-library"

PR #135, opened 2020-04-27
Add dateCreated to analytics app

ilopmar:


graemerocher:
This reminds, it is currently not using flyway and this change will break production tables. Could you add flyway support and a migration for this?

ilopmar:
Sure!

ilopmar:
@graemerocher I've added Flyway and created the initial migration. I've also added the `micronautVersion` to the table. I've tried locally and it works.

We need to define something like the following in the app to send the statistics:


```
---
micronaut:
  http:
    services:
      analytics:
        urls:
          - http://localhost:9090
```          
          
I've done it locally in `starter-web-netty` but we need probably need to define it in configuration in GCP for `starter-gcp-function`. I think we should use environment variables to keep the analitycs url hidden from users.

graemerocher:
It is already defined and working in production using an environment variable, so don't worry.

graemerocher:
Please fix the checkstyle violations

ilopmar:
Sorry, fixed.

Issue #130, opened 2020-04-25
Java CLI currently dies with a segfault


graemerocher:
Fails on all JDK versions. 

Changing the `projectVersion` in `gradle.properties` to `1.0.0.BUILD-SNAPSHOT` from `2.0.0.BUILD-SNAPSHOT` fixes the problem, however that is undesirable.

Example dump:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000000010e73327c, pid=55623, tid=10243
#
# JRE version: OpenJDK Runtime Environment GraalVM CE 20.0.0 (11.0.6+9) (build 11.0.6+9-jvmci-20.0-b02)
# Java VM: OpenJDK 64-Bit Server VM GraalVM CE 20.0.0 (11.0.6+9-jvmci-20.0-b02, mixed mode, sharing, tiered, jvmci, jvmci compiler, compressed oops, g1 gc, bsd-amd64)
# Problematic frame:
# V  [libjvm.dylib+0x53327c]  _ZN12LinkResolver21resolve_invokevirtualER8CallInfo6HandleRK18constantPoolHandleiP6Thread+0x4c
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/graemerocher/dev/micronaut/starter/hs_err_pid55623.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
```

graemerocher:
In addition we need an integration test for that actually tests the constructed CLI since the test suite currently doesn't pick up this issue

PR #140, opened 2020-04-27
Group features in categories

### LINKED ISSUE #66: Organize features in categories
Features should be organized into categories to help discovery

alvarosanchez:
Fixes #66.


CONTRIBUTING.md
```diff
@@ -20,9 +20,11 @@ To run the tests use `./gradlew check`.
 
 ### Running the CLI
 
-To generate a project execute `./gradlew run --args=""` where the args are what would be after `mn` using the CLI. For example: `./gradlew run --args="create-app temp"`. 
+To generate a project execute `./gradlew micronaut-cli:run --args=""` where the args are what would be after `mn` using the CLI. For example: `./gradlew micronaut-cli:run --args="create-app temp"`. 
 
-If you want to test output colors or the interactive shell, the jar must be executed. Run `./gradlew assemble` and then `java -jar starter-core/build/libs/starter-core-1.0.0.BUILD-SNAPSHOT.jar create-app temp`
+If you want to test output colors or the interactive shell, the jar must be executed. Run `./gradlew micronaut-cli:shadowJar` and then `java -jar starter-cli/build/libs/micronaut-cli-2.0.0.BUILD-SNAPSHOT-all.jar create-app temp`
+
+Note that interactive mode (ie: running without arguments) doesn't work when running the JAR [due to this issue](https://github.com/micronaut-projects/micronaut-starter/issues/130).
```
jameskleeh, line 27: This seems to be resolved?

alvarosanchez: Well, the problem described by #130 still persists when running with `java -jar` without arguments

jameskleeh: I've just pushed the fix for that problem, so this line can be removed

jameskleeh:
Looks great - please change Category to be a normal class with static strings and change the API to return a String. This is a more extensible approach and we aren't doing any comparisons of categories that would make an enum valuable.

alvarosanchez:
> Looks great - please change Category to be a normal class with static strings and change the API to return a String. That is a more extensible approach and we aren't doing any comparisons of categories that would make an enum valuable.

Done.

Issue #151, opened 2020-04-28
Add jOOQ formula


benmccann:
Wow I just found https://micronaut.io/start/ and can I say it's the coolest thing ever?! Way to go to the whole team that worked on this!

I'd love to see a formula for [Micronaut's jOOQ support](https://micronaut-projects.github.io/micronaut-sql/latest/guide/#jooq).

CC @lightoze @lukaseder as other people who might be interested in this

I may try to add it if I get time to figure out how. There isn't a guide on a adding a new formula by chance is there?

It looks like it maybe would go here:
* https://github.com/micronaut-projects/micronaut-starter/tree/master/starter-core/src/main/java/io/micronaut/starter/feature/database

graemerocher:
Yes that is right. We are still testing and plan to go live officially on Thursday. Thanks for the feedback!

lukaseder:
> CC @lightoze @lukaseder as other people who might be interested in this

Well, I definitely *am* interested in this! :-)

PR #153, opened 2020-04-28
Add jOOQ formula

benmccann:
Closes https://github.com/micronaut-projects/micronaut-starter/issues/151

@lightoze is there more we should add to this in a follow up PR? In any case, this seems like a good start to me

I have tests fail with this branch locally, but only ones that are failing on `master`

### DIFF
--- a/starter-core/src/test/groovy/io/micronaut/starter/feature/database/JooqSpec.groovy
+++ b/starter-core/src/test/groovy/io/micronaut/starter/feature/database/JooqSpec.groovy
@@ -1,4 +1,4 @@
-package io.micronaut.starter.feature.cassandra
+package io.micronaut.starter.feature.database
 
 import io.micronaut.starter.BeanContextSpec
 import io.micronaut.starter.application.ApplicationType
@@ -10,7 +10,7 @@ import spock.lang.Unroll
 class JooqSpec extends BeanContextSpec {
 
     @Unroll
-    void 'test gradle cassandra feature for language=#language'() {
+    void 'test gradle jooq feature for language=#language'() {
         when:
         String template = buildGradle.template(ApplicationType.DEFAULT, buildProject(), getFeatures(['jooq'], language)).render().toString()
 
@@ -22,7 +22,7 @@ class JooqSpec extends BeanContextSpec {
     }
 
     @Unroll
-    void 'test maven cassandra feature for language=#language'() {
+    void 'test maven jooq feature for language=#language'() {
         when:
         String template = pom.template(ApplicationType.DEFAULT, buildProject(), getFeatures(['jooq'], language), []).render().toString()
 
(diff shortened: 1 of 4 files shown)


starter-core/src/test/groovy/io/micronaut/starter/feature/database/JooqSpec.groovy
```diff
@@ -0,0 +1,42 @@
...
+
+        then:
+        template.contains('implementation("io.micronaut.configuration:micronaut-jooq")')
+
+        where:
+        language << Language.values().toList()
+    }
+
+    @Unroll
+    void 'test maven cassandra feature for language=#language'() {
```
jameskleeh, line 25: Fix test description

starter-core/src/test/groovy/io/micronaut/starter/feature/database/JooqSpec.groovy
```diff
@@ -0,0 +1,42 @@
...
+import io.micronaut.starter.application.ApplicationType
+import io.micronaut.starter.feature.build.gradle.templates.buildGradle
+import io.micronaut.starter.feature.build.maven.templates.pom
+import io.micronaut.starter.options.Language
+import spock.lang.Unroll
+
+class JooqSpec extends BeanContextSpec {
+
+    @Unroll
+    void 'test gradle cassandra feature for language=#language'() {
```
jameskleeh, line 13: Fix test description

starter-core/src/test/groovy/io/micronaut/starter/feature/database/JooqSpec.groovy
benmccann: done

starter-core/src/test/groovy/io/micronaut/starter/feature/database/JooqSpec.groovy
benmccann: done. thanks for catching those

graemerocher:
I think like the Hibernate feature this feature probably needs to apply the JDBC configuration and add H2 by default so it works out of the box otherwise the user will have to add configuration and manually pick a JDBC driver. See:

https://github.com/micronaut-projects/micronaut-starter/blob/master/starter-core/src/main/java/io/micronaut/starter/feature/database/HibernateJpa.java#L52-L62

benmccann:
Ok. I've added H2. It might be nice if there were the ability to specify options for features. I use Postgres so it's a little weird to add the H2 dependency to my project. We could have a dropdown that defaulted to H2, but let users choose Postgres or MySQL instead. It could maybe be under an advanced toggle if we want to keep things simple by default.

jameskleeh:
@benmccann The H2/Mysql/etc could easily be implemented. `--features mysql-driver` for example. You can submit an issue to track the request for that support

Issue #159, opened 2020-04-29
Add kotlin-runtime when Kotlin is selected


graemerocher:
See https://micronaut-projects.github.io/micronaut-kotlin/1.0.x/guide/#runtime

PR #175, opened 2020-04-29
fix: include kotlin-runtime from Kotlin

sdelamo:
Close: https://github.com/micronaut-projects/micronaut-starter/issues/159


jameskleeh:
@sdelamo It seems that kotlin-runtime does not have an API dependency on runtime, so its possible runtime would not be included  https://github.com/micronaut-projects/micronaut-kotlin/blob/master/kotlin-runtime/build.gradle

sdelamo:
I think it fails because io.micronaut.kotlin are not defined in core buildDependencies. See: https://github.com/micronaut-projects/micronaut-core/issues/3117

graemerocher:
@sdelamo this seems to be mixed with other changes could you rebase the PR with only the Kotlin changes

sdelamo:
> @sdelamo this seems to be mixed with other changes could you rebase the PR with only the Kotlin changes

I will do.

Issue #184, opened 2020-04-29
Handling of optional and related features


benmccann:
## Description

There are some features that would sometimes be useful, but perhaps not all users would want

E.g. in the GraphQL feature we wanted to demonstrate GraphiQL as being an optional feature

Another case where this came up was database drivers. Right now we're just automatically adding the H2 driver to database features like jOOQ and Hibernate, but it users might want to use Postgres, MySQL, etc. 

There are also cases where features may depend on each other. E.g. for jOOQ I'd like to add [gradle-jooq-plugin](https://github.com/etiennestuder/gradle-jooq-plugin). I don't know whether all users would want this, but I expect many or perhaps a majority would. This only makes sense if you have jOOQ though. Similarly for GraphQL many users would want to add [graphql-java-codegen-gradle](https://github.com/kobylynskyi/graphql-java-codegen/tree/master/plugins/gradle)

## Considerations

* How do we specify that things are related or depend on another? E.g. if you setup jOOQ or Hibernate you should have some database driver specified.
* How do we keep there some being too many features in the UI or keep them grouped together. E.g. the graphiql case is just a single line. It feels like it would clutter the UI to make a new feature for that unless it's grouped somehow

## Potential solutions

1. @jameskleeh suggested (in https://github.com/micronaut-projects/micronaut-starter/pull/153#issuecomment-621293216) a way of handling options such database drivers would be to have them as features (`--features mysql-driver` for example).
    1. If we handled them by adding features, would we have some kind of validation or suggestions to get you to add the other features needed? E.g. if you choose Hibernate maybe the UI description should suggest you also choose a database driver. Perhaps it chooses H2 for you automatically and you can change it? Similarly for GraphQL codegen or GraphiQL, you'd probably want the main GraphQL plugin. Or we could just make them show up in the UI next to each other so that it'd be hard to miss
2. One solution here for managing UI complexity could be subcategories. E.g. under GraphQL you could have graphql-code-gen and graphiql and under jooq you could have the jooq codegen plugin
    1. I like that this both adds organization and is relatively feasible to implement
3. Commented out code
[...]

PR #187, opened 2020-04-29
Create PostgreSQL feature

benmccann:
This refactoring sets the ground for allowing users to select other database types and follows the recommendation from https://github.com/micronaut-projects/micronaut-starter/issues/184#issuecomment-621390513

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/application/generator/GeneratorContext.java
+++ b/starter-core/src/main/java/io/micronaut/starter/application/generator/GeneratorContext.java
@@ -186,8 +186,11 @@ public class GeneratorContext {
     }
 
     public boolean isFeaturePresent(Class<? extends Feature> feature) {
-        Objects.requireNonNull(feature, "The feature class cannot be null");
         return features.isFeaturePresent(feature);
+    }
+
+    public <T extends Feature> T getFeature(Class<T> feature) {
+        return features.getFeature(feature);
     }
 
     public String getSourcePath(String path) {
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
@@ -443,8 +443,8 @@ List<Property> properties
 @if (features.contains("h2")) {
 @dependency.template("com.h2database", "h2", "runtime", null)
 }
-@if (features.contains("h2")) {
-@dependency.template("org.postgresql", "postgresql", "compile", "42.2.12")
+@if (features.contains("postgresql")) {
+@dependency.template("org.postgresql", "postgresql", "runtime", "42.2.12")
 }
 @if (features.contains("hibernate-gorm")) {
 @dependency.template("org.apache.tomcat", "tomcat-jdbc", "runtime", null)
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
@@ -443,6 +443,9 @@ List<Property> properties
 @if (features.contains("h2")) {
 @dependency.template("com.h2database", "h2", "runtime", null)
 }
+@if (features.contains("h2")) {
+@dependency.template("org.postgresql", "postgresql", "compile", "42.2.12")
+}
 @if (features.contains("hibernate-gorm")) {
 @dependency.template("org.apache.tomcat", "tomcat-jdbc", "runtime", null)
 }
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
@@ -443,7 +443,7 @@ List<Property> properties
 @if (features.contains("h2")) {
 @dependency.template("com.h2database", "h2", "runtime", null)
 }
-@if (features.contains("postgresql")) {
+@if (features.contains("postgres")) {
 @dependency.template("org.postgresql", "postgresql", "runtime", "42.2.12")
 }
 @if (features.contains("hibernate-gorm")) {
--- a/starter-core/src/main/java/io/micronaut/starter/feature/database/HibernateGorm.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/database/HibernateGorm.java
@@ -70,6 +70,14 @@ public class HibernateGorm implements LanguageSpecificFeature {
     @Override
     public void apply(GeneratorContext generatorContext) {
         Map<String, Object> config = generatorContext.getConfiguration();
+        DatabaseDriverFeature dbFeature = generatorContext.getFeature(DatabaseDriverFeature.class);
+        if (dbFeature == null) {
+            dbFeature = defaultDbFeature;
+        }
+        config.put("dataSource.url", dbFeature.getJdbcUrl());
+        config.put("dataSource.driverClassName", dbFeature.getDriverClass());
+        config.put("dataSource.username", dbFeature.getDefaultUser());
+        config.put("dataSource.password", dbFeature.getDefaultPassword());
         config.put("dataSource.pooled", true);
         config.put("dataSource.jmxExport", true);
         config.put("hibernate.hbm2ddl.auto", "update");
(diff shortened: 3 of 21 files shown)


benmccann:
It seems like the failing test is an unrelated flake. It's for Alexa and looks to be possibly related to a network failure. I wonder if there's anyway to turn off printing the name of every test that passes? It was really hard to find the failure in the logs because there was so much output

>io.micronaut.starter.generator.CreateAlexaSpec create-function-app with features [aws-alexa] java and maven and test framework: junit FAILED (2m)
  Condition not satisfied after 120.00 seconds and 118 attempts
      at spock.util.concurrent.PollingConditions.within(PollingConditions.java:170)
      at spock.util.concurrent.PollingConditions.eventually(PollingConditions.java:140)
      at io.micronaut.starter.generator.CommandSpec.testOutputContains(CommandSpec.groovy:53)
      at io.micronaut.starter.generator.CreateAlexaSpec.create-function-app with features #features #lang and #build and test framework: #testFramework(CreateAlexaSpec.groovy:55)
  Caused by: Condition not satisfied:
  output.toString().contains(value)
  |      |          |        |
  |      |          false    BUILD SUCCESS
  |      [INFO] Scanning for projects...
  |      [INFO] 
  |      [INFO] -----------------------< example.micronaut:foo >------------------------
  |      [INFO] Building foo 0.1
  |      [INFO] --------------------------------[ jar ]---------------------------------
  |      Downloading from jcenter.bintray.com: https://jcenter.bintray.com/io/micronaut/aws/micronaut-function-aws-alexa/2.0.0.M2/micronaut-function-aws-alexa-2.0.0.M2.pom
  |      Progress (1): 0.8/3.3 kB
  |      Progress (1): 2.2/3.3 kB
  |      Progress (1): 3.3 kB

starter-core/src/main/java/io/micronaut/starter/feature/database/HibernateGorm.java
```diff
@@ -70,12 +70,8 @@ public void processSelectedFeatures(FeatureContext featureContext) {
     @Override
     public void apply(GeneratorContext generatorContext) {
         Map<String, Object> config = generatorContext.getConfiguration();
-        config.put("dataSource.url", "jdbc:h2:mem:devDb;MVCC=TRUE;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE");
         config.put("dataSource.pooled", true);
         config.put("dataSource.jmxExport", true);
-        config.put("dataSource.driverClassName", "org.h2.Driver");
-        config.put("dataSource.username", "sa");
-        config.put("dataSource.password", "");
```
jameskleeh, line 78: This will definitely cause GORM applications to not function. We can't assume because the H2 feature is added that it is using `micronaut-jdbc` and expecting `datasources.default`. I think you'll need to leave the configuration up to the individual feature, however perhaps add a method or methods on the driver feature to make that easier.

starter-core/src/test/groovy/io/micronaut/starter/feature/database/HibernateGormSpec.groovy
```diff
@@ -71,7 +71,7 @@ class HibernateGormSpec extends BeanContextSpec {
         GeneratorContext ctx = buildGeneratorContext(['hibernate-gorm'])
 
         then:
-        ctx.configuration.containsKey("dataSource.url")
+        ctx.configuration.containsKey("datasources.default.url")
```
jameskleeh, line 74: This is incorrect

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
```diff
@@ -434,6 +434,9 @@ List<Property> properties
 @if (features.contains("h2")) {
 @dependency.template("com.h2database", "h2", "runtime", null)
 }
+@if (features.contains("h2")) {
```
jameskleeh, line 437: This should be `.contains(postgressql)`

starter-core/src/test/groovy/io/micronaut/starter/feature/database/HibernateGormSpec.groovy
benmccann: fixed

starter-core/src/main/java/io/micronaut/starter/feature/database/HibernateGorm.java
benmccann: Fixed

jameskleeh: This config is still being removed, so this will still cause GORM applications to not function

starter-core/src/test/groovy/io/micronaut/starter/feature/database/HibernateGormSpec.groovy
```diff
@@ -71,7 +71,7 @@ class HibernateGormSpec extends BeanContextSpec {
         GeneratorContext ctx = buildGeneratorContext(['hibernate-gorm'])
 
         then:
-        ctx.configuration.containsKey("dataSource.url")
+        ctx.configuration.containsKey("datasource.url")
```
jameskleeh, line 74: This is incorrect

starter-core/src/main/java/io/micronaut/starter/application/generator/GeneratorContext.java
```diff
@@ -186,10 +186,13 @@ public void applyFeatures() {
     }
 
     public boolean isFeaturePresent(Class<? extends Feature> feature) {
-        Objects.requireNonNull(feature, "The feature class cannot be null");
```
jameskleeh, line 189: Why was this removed?

starter-core/src/main/java/io/micronaut/starter/feature/database/HibernateGorm.java
benmccann: fixed

starter-core/src/test/groovy/io/micronaut/starter/feature/database/HibernateGormSpec.groovy
benmccann: oops. fixed

starter-core/src/main/java/io/micronaut/starter/application/generator/GeneratorContext.java
benmccann: `features.isFeaturePresent` has the same check, so it was redundant

starter-core/src/main/java/io/micronaut/starter/feature/database/HibernateGorm.java
```diff
@@ -70,12 +70,16 @@ public void processSelectedFeatures(FeatureContext featureContext) {
     @Override
     public void apply(GeneratorContext generatorContext) {
         Map<String, Object> config = generatorContext.getConfiguration();
-        config.put("dataSource.url", "jdbc:h2:mem:devDb;MVCC=TRUE;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE");
+        DatabaseDriverFeature dbFeature = generatorContext.getFeature(DatabaseDriverFeature.class);
+        if (dbFeature == null) {
+            dbFeature = defaultDbFeature;
+        }
```
jameskleeh, line 76: The `getFeature` method should return an optional, with a `getRequiredFeature` that returns the feature and throws an exception if it doesn't exist. In this scenario we should expect that the feature exists.

benmccann: Yeah, I had something close to that at first. The difficulty I ran into was that `FeatureSpec` calls `applyFeatures` on every feature without first calling `processSelectedFeatures`, so the database features aren't present and it would result in an exception.

Do you have any thoughts on how to handle that? Should I try to add a database feature always in `FeatureSpec`? That might not make sense for some features, but it might be messy to try to detect which features are the ones that require it and then add it. Or should I just leave the `null` handling here?

jameskleeh: @benmccann The line in the test should probably be `getFeatures([feature.name]).getFeatures()` then

benmccann: Sorry, I'm not following. Are you suggesting I go ahead and add `H2` to the `generatorContext` of for every feature in the test whether or not it's DB related?

benmccann: @jameskleeh thanks for the review! I'd love to close this out. Any advice on moving forward?

PR #219, opened 2020-05-08
[starter-web-netty]  CORS_ALLOWED_ORIGIN config property

eahrold:
This helps with developing the web ui against the starter.

### DIFF
--- a/starter-web-netty/src/main/resources/application.yml
+++ b/starter-web-netty/src/main/resources/application.yml
@@ -10,7 +10,7 @@ micronaut:
       configurations:
         web:
           allowedOrigins:
-            - ^http(|s):\/\/micronaut\.io$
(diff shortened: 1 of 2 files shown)


starter-web-netty/src/main/resources/application.yml
```diff
@@ -10,7 +10,7 @@ micronaut:
       configurations:
         web:
           allowedOrigins:
-            - ^http(|s):\/\/micronaut\.io$
+            - ${CORS_ALLOWED_ORIGIN:^http(|s):\/\/micronaut\.io$}
```
ilopmar, line 13: you need to escape the default value with backticks becase there is a colon:
I think it should be ${CORS_ALLOWED_ORIGIN:`^http(|s):\/\/micronaut\.io$`}

eahrold: Good catch @ilopmar, thanks. Addressed b3971f0

Issue #249, opened 2020-05-25
Add acme to set of features


sdelamo:
See: https://micronaut-projects.github.io/micronaut-acme/snapshot/guide/

jameskleeh:
Duplicate of https://github.com/micronaut-projects/micronaut-starter/issues/53

PR #251, opened 2020-05-26
Micronaut acme support

### LINKED ISSUE #53: create acme related commands
After talking with @jameskleeh we decided to move the acme-cli that was built here into this project.

https://github.com/micronaut-projects/micronaut-acme/tree/master/acme-cli

Current state of the acme-cli is it helps you create keys and create/deactivate accounts. In its current form. Probably other features that could be added but that is where it will start.

zendern:
I believe this will fix #53 and #249. 

What still needs worked on is there are still a handful of failing tests that need to be address and some manual testing of the cli using the feature still needs to be done. 

I did add the acme4j jar in what I believe is the right location and since i copied the tests over from the previous acme-cli that means I also added testcontainers at testImplementation time.

### DIFF
--- a/starter-cli/src/test/groovy/io/micronaut/starter/cli/feature/acme/CliBaseSpec.groovy
+++ b/starter-cli/src/test/groovy/io/micronaut/starter/cli/feature/acme/CliBaseSpec.groovy
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.micronaut.starter.cli.feature.acme
+
+import io.micronaut.context.BeanContext
+import io.micronaut.core.io.socket.SocketUtils
+import io.micronaut.starter.cli.CommandFixture
+import io.micronaut.starter.cli.CommandSpec
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.testcontainers.Testcontainers
+import org.testcontainers.containers.GenericContainer
+import org.testcontainers.containers.wait.strategy.HostPortWaitStrategy
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy
+import org.testcontainers.containers.wait.strategy.WaitAllStrategy
+import org.testcontainers.utility.MountableFile
+import spock.lang.AutoCleanup
+import spock.lang.Ignore
+import spock.lang.Shared
+
+import java.time.Duration
+
+@Ignore
+class CliBaseSpec extends CommandSpec implements CommandFixture {
+    private static final Logger log = LoggerFactory.getLogger(CliBaseSpec.class)
+
+    @Shared
+    @AutoCleanup
+    BeanContext beanContext = BeanContext.run()
+
+    @Shared
+    int expectedHttpPort
+
+    @Shared
+    int expectedSecurePort
+
+    @Shared
+    int expectedPebbleServerPort
+
+    @Shared
+    GenericContainer certServerContainer
+
+    @Shared
+    String acmeServerUrl
+
+    def setupSpec(){
+        expectedHttpPort = SocketUtils.findAvailableTcpPort()
+        expectedSecurePort = SocketUtils.findAvailableTcpPort()
+        expectedPebbleServerPort = SocketUtils.findAvailableTcpPort()
+
+        Testcontainers.exposeHostPorts(expectedHttpPort, expectedSecurePort)
+
+        def file = File.createTempFile("pebble", "config")
+        file.write """{
+              "pebble": {
+                "listenAddress": "0.0.0.0:${expectedPebbleServerPort}",
+                "certificate": "test/certs/localhost/cert.pem",
+                "privateKey": "test/certs/localhost/key.pem",
+                "httpPort": $expectedHttpPort,
+                "tlsPort": $expectedSecurePort
+              }
+            }"""
+
+        log.info("Expected micronaut ports - http : {}, secure : {} ", expectedHttpPort, expectedSecurePort)
+        log.info("Expected pebble config : {}", file.text)
+
+        certServerContainer = new GenericContainer("letsencrypt/pebble:latest")
+                .withCopyFileToContainer(MountableFile.forHostPath(file.toPath()), "/test/config/pebble-config.json")
+                .withCommand("/usr/bin/pebble", "-strict", "false")
+                .withEnv(getPebbleEnv())
+                .withExposedPorts(expectedPebbleServerPort)
+                .waitingFor(new WaitAllStrategy().withStrategy(new LogMessageWaitStrategy().withRegEx(".*ACME directory available.*\n"))
+                        .withStrategy(new HostPortWaitStrategy())
+                        .withStartupTimeout(Duration.ofMinutes(2)));
+        certServerContainer.start()
+
+        acmeServerUrl = "acme://pebble/${certServerContainer.containerIpAddress}:${certServerContainer.getMappedPort(expectedPebbleServerPort)}"
+    }
+
+    Map<String, String> getPebbleEnv(){
+        return [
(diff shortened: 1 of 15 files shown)


graemerocher:
@zendern Thanks for the PR, @jameskleeh will you help @zendern get this one over the line

starter-core/src/main/java/io/micronaut/starter/feature/acme/Acme.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package io.micronaut.starter.feature.acme;
+
+import io.micronaut.starter.application.generator.GeneratorContext;
+import io.micronaut.starter.feature.server.ServerFeature;
+
+import javax.inject.Singleton;
+
+@Singleton
+public class Acme implements ServerFeature {
```
zendern, line 24: This felt like the right Feature type to me but maybe it belongs elsewhere?? Other?? Maybes its own Feature set. Looking or suggestions here.

jameskleeh: It shouldn't implement ServerFeature. Its not a server. The category the feature belongs in and the interface it implements aren't necessarily related. I think this belongs in other personally

zendern: That works I’ll go ahead and move it to other

zendern: Moved over to other in commit c5e92e1. But now that I moved it over to `OTHER` I get a failing test. 

image

This tests makes it seem that a feature should never be categorized as `OTHER` if I understand it correctly. I can change that test or maybe I misunderstand?? @jameskleeh any  thoughts??

If I do a `mn create-app --list-features` it looks like I would expect
image

jameskleeh: Not really sure what the best choice is. You can create a new category (SSL) or something like that. It does fit into Server, however I don't want it to be confused with the other features that are actually servers.

zendern: Alright I went with `SSL` ¯\_(ツ)_/¯ felt like `Security` could have been a home but it's kind of a similar vain as `Server`, it fits but doesn't really fit with all the other tools. Considered `Encryption` and even `Certificates` but just settled with `SSL`. Still open for suggestions. If you happen to know or can think of some other things that might fit the category in the future maybe one of those makes more sense over `SSL`.

starter-cli/src/test/groovy/io/micronaut/starter/cli/feature/acme/CliBaseSpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy
+import org.testcontainers.containers.wait.strategy.WaitAllStrategy
+import org.testcontainers.utility.MountableFile
+import spock.lang.AutoCleanup
+import spock.lang.Ignore
+import spock.lang.Shared
+
+import java.time.Duration
+
+@Ignore
```
jameskleeh, line 37: This should probably be an abstract class so `@Ignore` is not required

starter-core/src/main/java/io/micronaut/starter/feature/acme/Acme.java
```diff
@@ -0,0 +1,69 @@
...
+
+import io.micronaut.starter.application.ApplicationType;
+import io.micronaut.starter.application.generator.GeneratorContext;
+import io.micronaut.starter.feature.Category;
+import io.micronaut.starter.feature.OneOfFeature;
+
+import javax.inject.Singleton;
+
+@Singleton
+public class Acme implements OneOfFeature {
```
jameskleeh, line 26: This should not implement OneOfFeature. There is only 1 acme feature so there can only be 1. One of is for ensuring only 1 database driver can be included for example.

starter-core/src/main/java/io/micronaut/starter/feature/acme/Acme.java
```diff
@@ -0,0 +1,69 @@
...
+        return "https://micronaut-projects.github.io/micronaut-acme/latest/guide/index.html";
+    }
+
+    @Override
+    public Class<?> getFeatureClass() {
+        return Acme.class;
+    }
+
+    @Override
+    public boolean supports(ApplicationType applicationType) {
```
jameskleeh, line 61: I don't think this is applicable to all application types. I think this would only apply to the default application type.

starter-core/src/main/java/io/micronaut/starter/feature/acme/Acme.java
```diff
@@ -0,0 +1,69 @@
...
+        return "ACME";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Adds support for ACME (Automated Certificate Management Environment)";
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
```
jameskleeh, line 44: This should require the netty server implementation because the library is designed with that in mind. If someone chooses `jetty` and `acme` it will not produce the desired outcome. This can be achieved by creating a `FeatureValidator`

starter-cli/src/test/groovy/io/micronaut/starter/cli/feature/acme/CliBaseSpec.groovy
zendern: Cleaned up in 0839f10

starter-core/src/main/java/io/micronaut/starter/feature/acme/Acme.java
zendern: Fixed in 97df988

starter-core/src/main/java/io/micronaut/starter/feature/acme/Acme.java
zendern: Fixed in 97df988

starter-core/src/main/java/io/micronaut/starter/feature/acme/Acme.java
zendern: Validator added here 14f4a4e

zendern:
I also opened this up as well since it will be needed for this to actually work when someone adds the feature to their project.

https://github.com/micronaut-projects/micronaut-core/pull/3497

graemerocher:
Thanks for the contribution!

graemerocher:
@zendern this may need updating before releasing 1.0.0 https://micronaut-projects.github.io/micronaut-acme/snapshot/guide/#installation

We plan to release another RC on Monday so it would be good if you could release 1.0.0 in time for that.

zendern:
> @zendern this may need updating before releasing 1.0.0 https://micronaut-projects.github.io/micronaut-acme/snapshot/guide/#installation
> 
> 
> 
> We plan to release another RC on Monday so it would be good if you could release 1.0.0 in time for that.

It's on my planned Todo list for this weekend so should be ready for Monday's cut...I'll keep you posted 👍

graemerocher:
@zendern Great. One more thing, if you could test the native build just in case you spot any issues with the native executable.

Mac: https://github.com/micronaut-projects/micronaut-starter/actions/runs/140897770
Linux: https://github.com/micronaut-projects/micronaut-starter/actions/runs/140897910

zendern:
@graemerocher in regards to the native builds. Yeah very broken. 😬  Looking into it now but for some reason all the flag/options are not present so I cant even use them. Is this something you have seen before?? 

```
zender [Downloads/mn-darwin-amd64-snapshot/hello-world] → ../bin/mn create-acme-account -h
Usage: mn create-acme-account [-fhvVx]
Creates a new account on the given ACME server
  -f, --force        Whether to overwrite existing files
  -h, --help         Show this help message and exit.
  -v, --verbose      Create verbose output.
  -V, --version      Print version information and exit.
  -x, --stacktrace   Show full stack trace when exceptions occur.
```

What it should and does look like with the non native version.

```
zender [distributions/micronaut-cli-2.0.0.BUILD-SNAPSHOT/hello-world]  master ✔ ↑ → ../bin/mn create-key                                          [f1a4c56e]
Missing required option '--key-name=<keyName>'
Usage: mn create-key [-fhvVx] [-k=<keyDir>] -n=<keyName> [-s=<keySize>]
Creates an keypair for use with ACME integration
  -f, --force                Whether to overwrite existing files
  -h, --help                 Show this help message and exit.
  -k, --key-dir=<keyDir>     Custom location on disk to put the key to be used
                               with this account.
                               Default: src/main/resources
  -n, --key-name=<keyName>   Name of the key to be created
  -s, --key-size=<keySize>   Size of the key to be generated
                               Default: 4096
  -v, --verbose              Create verbose output.
  -V, --version              Print version information and exit.
  -x, --stacktrace           Show full stack trace when exceptions occur.
```

PR #266, opened 2020-06-05
Testcontainers

jameskleeh:


starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
```diff
@@ -509,6 +509,23 @@ dependencies {
 @if (features.contains("neo4j-bolt")) {
     @dependency.template("org.neo4j.test", "neo4j-harness", "testRuntime", null)
 }
+@if (features.contains("testcontainers")) {
+@if (features.contains("mysql")) {
+    @dependency.template("org.testcontainers", "mysql", "testRuntimeOnly", "1.14.3")
```
graemerocher, line 514: I think we should add test containers to the BOM rather than hard coding the version

starter-core/src/main/java/io/micronaut/starter/feature/database/DatabaseDriverConfigurationFeature.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.starter.feature.database;
+
+import io.micronaut.starter.feature.Feature;
+
+public interface DatabaseDriverConfigurationFeature extends Feature {
```
graemerocher, line 20: some javadoc on this one would be good

graemerocher:
One other thing, the tests don't seem to be running on the CI? https://github.com/micronaut-projects/micronaut-starter/runs/743604628?check_suite_focus=true#step:5:85

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
jameskleeh: Because we have no code that relies on testcontainers and thus do not need to ensure the correct version is used, I don't think it should be in the bom. If it were, the only purpose it would serve is to leave off the version in the build, which is not the primary purpose of the bom.

jameskleeh:
@graemerocher Seems our cache is a little over zealous... https://github.com/micronaut-projects/micronaut-starter/runs/743604615?check_suite_focus=true#step:5:72

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
alvarosanchez: It may not be the primary purpose of a BOM, but hardcoding versions in strings within a Rocker template will make them go outdated and unnoticed pretty quickly.

graemerocher: Actually this is a perfect example of a use case for the BOM. The BOM is so that we can express what the minimum supported version is, if users then want to override that with another version that is up to them otherwise we leave it unexpressed what versions of Testcontainers we support which in fact could be a problem for this feature because only newer versions have the JDBC URL feature.

Apart from that as @alvarosanchez if we hard code this then it is likely it will never get updated by anyone until something breaks.

jameskleeh: > The BOM is so that we can express what the minimum supported version is

This isn't the minimum supported version though. I don't know what the minimum version is. If this logic is sound then we should have no versions whatsoever in the entire build and we should be adding the following to the BOM. I disagree with using the BOM for this purpose

```
    @dependency.template("de.flapdoodle.embed", "de.flapdoodle.embed.mongo", "testImplementation", "2.0.1")
    @dependency.template("com.microsoft.sqlserver", "mssql-jdbc", "runtimeOnly", "8.2.2.jre8")
    @dependency.template("org.postgresql", "postgresql", "runtimeOnly", "42.2.12")
    @dependency.template("mysql", "mysql-connector-java", "runtimeOnly", "8.0.20")
    @dependency.template("org.mariadb.jdbc", "mariadb-java-client", "runtimeOnly", "2.6.0")
    @dependency.template("com.oracle.ojdbc", "ojdbc8", "runtimeOnly", "19.3.0.0")
    @dependency.template("org.apache.logging.log4j", "log4j-core", "implementation", "2.12.1")
    @dependency.template("org.apache.logging.log4j", "log4j-api", "runtimeOnly", "2.12.1")
    @dependency.template("org.apache.logging.log4j", "log4j-slf4j-impl", "runtimeOnly", "2.12.1")
    @dependency.template("ch.qos.logback", "logback-classic", "runtimeOnly", "1.2.3")
    @dependency.template("com.fasterxml.jackson.module", "jackson-module-kotlin", "runtimeOnly", "2.9.8")
    @dependency.template("io.ktor", "ktor-server-netty", "implementation", "1.3.2")
    @dependency.template("io.ktor", "ktor-jackson", "implementation", "1.3.2")
```

graemerocher: All of those should indeed also go into the BOM, I don't know what the big disagreement is, we have always used the BOM for this purpose even in Grails where we avoided putting hard coded versions in the profiles.

Also our BOM has a lot fewer dependencies in it compared to Spring which has pretty much every third party dependency in there. 

The point of a BOM is describe what dependencies play well together rather than leaving that up to chance and the hope that they will work together.

Can I ask what are the negative aspects that you believe users will experience from having test containers (or any other third party dependency) in the BOM?

jameskleeh: From my perspective the negative aspects are:

1. It's not clear to the user what version is being used. They have to run some dependency report, etc to know.
2. The potential difficulty of overriding the version depending on the platform option
3. A philosophical idea that Micronaut shouldn't tell you what version to use of any library that isn't directly tied to what Micronaut is offering. I don't think a feature in the starter applies here.
4. If the BOM is forever added to then it will become like Spring's in the future and will take more and more developer time to maintain it / triage issues related to it / update and test.

jameskleeh: As it stands now, if we upgrade the version of testcontainers, we will immediately know whether or not it works. If it is moved to the BOM, then we have to wait for the snapshot to update then a build on this project to run to know if it works. We already have enough of that with our own projects.

graemerocher: 1. The version they are using is the supported version by Micronaut
2. This is only an issue with Gradle and only when using `enforcedPlatform` which it appears is no longer necessary with the correct latest version checking in Gradle
3. It is not about telling it is about specifying what the compatible set of dependencies are rather than leaving that up to chance
4. Is that a bad thing? I have already had more than one user at conferences come to me to say one of the reasons they use Spring Boot is the defined set of compatibility dependencies and not have to figure out the right versions to use of everything. They regard it a negative that you have to go to search.maven.org to figure out the version of certain dependencies in order to add it to your project. I don't believe that that is a great user experience.
5. It is not our job to make our lives easier, it is out job to make our users lives easier. The fact that it is more difficult for us to test is neither here nor there.

jameskleeh: 1. That isn't a valid response. The "supported version by Micronaut" is not "the version".
2. Right likely not an issue, but definitely still needs to be kept in mind
3. If the version is specified, its not up to chance
4. I guess this is a matter of personal preference
5. Clearly the maintainability of a library is important. Time spent doing unnecessary maintenance tasks is time that could have been spent fixing bugs or adding new features.

The feedback cycle of updates to actual tests against that update is singularly enough of a reason for this particular case to keep the version in the build in my opinion.

graemerocher: Well I guess we will have to agree to disagree

sdelamo:
I've created an issue around creating a single place to define version numbers exposed by features: 

https://github.com/micronaut-projects/micronaut-starter/issues/270

jameskleeh:
@graemerocher The bom has been added. Ready for review

Issue #261, opened 2020-05-31
Replace Kotlin Test with Kotest


ilopmar:
According to https://micronaut-projects.github.io/micronaut-test/snapshot/guide/index.html#_1_2_x micronaut-test snapshot add supports for Kotest and deprecates support for Kotlin Test.

We should replace[ Kotlin test](https://github.com/kotest/kotest/blob/master/doc/reference_3.3.md) with [KoTest](https://github.com/kotest/kotest).

PR #273, opened 2020-06-12
Temporarily use JUnit by default with Kotlin

ilopmar:
I've preferred to ignore one test instead of "fixing" it the wrong way, because this commit is only temporarily. Once we start working on #261 we need to revert this commit.


jameskleeh:
@sdelamo Do you know why the aws test might be failing? I can't reproduce locally

graemerocher:
I couldn't reproduce either, merging for now

Issue #276, opened 2020-06-15
Feature Spring-based transaction management


sdelamo:
`implementation("io.micronaut.sql:micronaut-hibernate-jpa-spring")`

It enables configuration of the Spring transaction management.

PR #277, opened 2020-06-15
build: add spring transaction management feature

sdelamo:
Close: https://github.com/micronaut-projects/micronaut-starter/issues/276

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
@@ -438,6 +438,9 @@ dependencies {
 @if (features.contains("spring")) {
     @dependency.template("org.springframework.boot", "spring-boot-starter", "implementation", null)
 }
+@if (features.contains("spring-tx")) {
+    @dependency.template("io.micronaut.sql", "micronaut-hibernate-jpa-spring", "implementation", null)
+}
 @if (features.contains("spring-web")) {
     @dependency.template("org.springframework.boot", "spring-boot-starter-web", "implementation", null)
     @dependency.template("io.micronaut", "micronaut-http-server", "implementation", null)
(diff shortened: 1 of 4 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
```diff
@@ -438,6 +438,9 @@ dependencies {
 @if (features.contains("spring")) {
     @dependency.template("org.springframework.boot", "spring-boot-starter", "implementation", null)
 }
+@if (features.contains("spring-tx")) {
+    @dependency.template("io.micronaut.sql", "micronaut-hibernate-jpa-spring", "implementation", null)
```
graemerocher, line 442: This doesn't make any sense to me. You can use Spring transactions with just JDBC or just Micronaut Data JDBC so it has nothing to do with `micronaut-hibernate-jpa-spring`

If we are going to merge this PR then the feature should also check if Hibernate is selected to apply this and should also work with just JDBC

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
```diff
@@ -436,6 +436,9 @@ List<Property> properties
 @dependency.template("io.micronaut.spring", "micronaut-spring-annotation", "compile", null)
 }
 }
+@if (features.contains("spring-tx")) {
+@dependency.template("io.micronaut.sql", "micronaut-hibernate-jpa-spring", "compile", null)
```
graemerocher, line 440: Same comment as above

starter-core/src/main/java/io/micronaut/starter/feature/spring/SpringTx.java
```diff
@@ -0,0 +1,59 @@
...
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.micronaut.starter.application.ApplicationType;
+import io.micronaut.starter.feature.Category;
+import io.micronaut.starter.feature.Feature;
+
+import javax.inject.Singleton;
+
+@Singleton
+public class SpringTx implements Feature {
```
graemerocher, line 26: Users should be using the Spring annotations to define transactions when this feature is selected that means this feature also needs to apply the Spring feature when selected since `micronaut-spring-annotation` needs to be on the annotation processor classpath. 

See https://github.com/micronaut-projects/micronaut-spring/issues/73

alvarosanchez:
Is this going to be moved forward?

PR #283, opened 2020-06-18
Use SDKMAN! 2.0 plugin and set sdkMajorRelease

graemerocher:


.github/workflows/release.yml
```diff
@@ -216,7 +216,7 @@ jobs:
         env:
           GVM_SDKVENDOR_KEY: ${{ secrets.GVM_SDKVENDOR_KEY }}
           GVM_SDKVENDOR_TOKEN: ${{ secrets.GVM_SDKVENDOR_TOKEN }}
-        run: ./gradlew sdkMinorRelease --no-daemon
+        run: ./gradlew sdkMajorRelease --no-daemon
```
marc0der, line 219: Not sure if this is intentional, but will now no longer set the release version as the candidate default.

PR #287, opened 2020-06-21
Fix Native Image issues for new ACME commands

zendern:
I was missing 2 things 

1. @ReflectiveAccess on all the @CommandLine.Option so they just did not show up at all
2. The AcmeServerOption was not being picked up correctly so needed to add a constructor and add it to @TypeHint for graal to be happy with it. 

I also added a test to help catch future @ReflectiveAccess issues if they are not added in new commands.

### DIFF
--- a/starter-cli/build.gradle
+++ b/starter-cli/build.gradle
@@ -41,6 +41,7 @@ dependencies {
     generateConfig "info.picocli:picocli-codegen:${picocliVersion}"
 
     testImplementation "org.testcontainers:spock:1.13.0"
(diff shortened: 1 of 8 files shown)


zendern:
@graemerocher this should fix up some native image issues that i found with the new ACME commands. I also cut v1.0.0 of micronaut-acme. 

https://github.com/micronaut-projects/micronaut-acme/releases/tag/v1.0.0

PR for bom is here too
https://github.com/micronaut-projects/micronaut-core/pull/3536

So hopefully that should button up everything for your RC cut for tomorrow.

PR #296, opened 2020-06-29
Generate resource-config.json for OpenAPI html if GraalVM is selected

ilopmar:


graemerocher:
I think this should go into the `GraalTypeElementVisitor` which is already generating resource config

ilopmar:
The problem is that the swagger files are generated by the OpenAPI visitors in `build/classes/java/main/META-INF/swagger` and not in `src/main/resources` hence the `GraalTypeElementVisitor` can't see those files to add them automatically.

graemerocher:
@ilopmar yes but it can see if the Swagger dependencies are on the classpath and simply generate the config if it is

PR #330, opened 2020-07-11
Generated Maven pom.xml use defaults from Micronaut parent pom

alejandropg:
Whit this PR I try to use the defaults defined in the Micronaut parent pom (my previous PRs in micronaut-core project) so the generated poms by the starter are simpler and more "Maven compliant",


alejandropg:
WIP, don't merge!

PR #332, opened 2020-07-12
Use Maven with JUnit default tests names

alejandropg:
Right now there is a problem with the Maven pom.xml generated by the starter when you use JUnit. The problem is that Surefire (the Maven plugin that runs unit tests) is configured to include tests in any file `<include>%regex[.*]</include>`. **This cause that, in unit testing phase, Surefire tries to run also integration tests, and build fails.**

In Maven the expected default is that Surefire runs tests in files like:
  - **/Test*.java
  - **/*Test.java
  - **/*Tests.java
  - **/*TestCase.java
and Failsafe plugin runs integration tests in files like:
  - **/IT*.java
  - **/*IT.java
  - **/*ITCase.java

With this commit the Maven pom.xml is generated with the espected defaults and is only reconfigured if Spock or Kotest are used. And in wich case the espected Spock or Kotest default filenames are used.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
@@ -583,32 +583,27 @@ List<Property> properties
         <artifactId>maven-shade-plugin</artifactId>
       </plugin>
 }
+@if (features.testFramework().isKotlinTest() || features.testFramework().isSpock()) {
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-@if (features.testFramework().isJunit() || features.testFramework().isSpock()) {
-        <configuration>
-          <detail>true</detail>
-          <includes>
-            <include>%regex[.*]</include>
-          </includes>
-        </configuration>
-} else if (features.testFramework().isKotlinTest()) {
         <configuration>
           <includes>
             <include>**/*Spec.*</include>
             <include>**/*Test.*</include>
           </includes>
         </configuration>
+@if (features.testFramework().isKotlinTest()) {
         <dependencies>
           <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
-            <version>5.3.1</version>
+            <version>${junit5.version}</version>
           </dependency>
         </dependencies>
 }
       </plugin>
+}
       <plugin>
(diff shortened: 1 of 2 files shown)


alvarosanchez:
Looks good to me, thanks!

Could it be possible to add some tests so that we can make sure it doesn't break in the future?

alejandropg:
Sure, no problem.

But these tests are a little "strange" because I have to test the absence of something ([I already did something similar in my previous commit](https://github.com/micronaut-projects/micronaut-starter/blob/d651142888f4bf38c5a51bd6eea65cd65c8e861f/starter-core/src/test/groovy/io/micronaut/starter/feature/build/maven/MavenSpec.groovy#L27-L31))

If you want I can do the same. If is the case I need help to know where to put the tests, maybe in <https://github.com/micronaut-projects/micronaut-starter/blob/fc1eba657695431dd5e78ea136fdc87efeb46de6/starter-core/src/test/groovy/io/micronaut/starter/feature/test/JUnitSpec.groovy>? But this file looks like "only Gradle", maybe could be better to create a new file and rename the actual one, some thing like `JUnitGradleSpec` and` JUnitMavenSpec`

Please, tell me what do you prefer and I will try.

alvarosanchez:
> But these tests are a little "strange" because I have to test the absence of something (I already did something similar in my previous commit)

That's the way to go. If it becomes present again, it should fail.

> But this file looks like "only Gradle", maybe could be better to create a new file and rename the actual one, some thing like JUnitGradleSpec and JUnitMavenSpec

Don't have a strong opinion on that. In general we tend to group spec classes around features. See https://github.com/micronaut-projects/micronaut-starter/blob/master/starter-core/src/test/groovy/io/micronaut/starter/feature/database/OracleSpec.groovy as an example of something that has both Maven and Gradle tests.

alejandropg:
Ok with one single test file for Maven and Gradle. I asked because I didn't know your style, but this is perfect 👌

alejandropg:
@alvarosanchez please check if these tests are what you are looking for.
Finally I didn't add them to `JUnitSpec` because really they are not related with JUnit. The tests are about parent defaults for Surefire and Failsafe that are just the "launcher" but could work with several "runners" (JUnit, TestNG, ...)
Anyway if you prefer other layout, tell me and I could move them where do you want.

alvarosanchez:
I was expecting to see a test of what would happen if you were to choose Kotlin Test or Spock, since those are the conditionals you added. But don't want to be too picky, that can be done later. This looks good to me as is. Thanks!

alejandropg:
Ok don't worry I will make in another PR, I promise 🤞😉
Thanks!

PR #347, opened 2020-07-19
Add tests of Maven configuration for unit tests with Spock or KotlinTest

alejandropg:
Add tests to check the generated pom.xml when you are using Spock or KotlinTest.

@alvarosanchez I think this is what your was looking for. Tell me If you want that I change something.

### DIFF
--- a/starter-core/src/test/groovy/io/micronaut/starter/feature/test/KotlinTestSpec.groovy
+++ b/starter-core/src/test/groovy/io/micronaut/starter/feature/test/KotlinTestSpec.groovy
@@ -1,0 +1,41 @@
+package io.micronaut.starter.feature.test
+
+import io.micronaut.starter.BeanContextSpec
+import io.micronaut.starter.application.ApplicationType
+import io.micronaut.starter.feature.Features
+import io.micronaut.starter.feature.build.maven.templates.pom
+import io.micronaut.starter.options.BuildTool
+import io.micronaut.starter.options.TestFramework
+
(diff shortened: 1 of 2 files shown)


alvarosanchez:
Thanks!

alejandropg:
@alvarosanchez I think with this I finished my proposals for now. Any way if you want to suggest something to me, don't hesitate and contact me 👍
Thanks a lot, and really congrats! 👏👏👏👏

graemerocher:
Thanks for all the contributions, big improvements. Wish there was a way to move the annotation processor config to the parent, but couldn't work out myself how that was possible.

alejandropg:
@graemerocher the problem is that the parent is "static" so you can not change nor configure it depending of what are you using in your project.

Our approximation was to create two parent, one for Java and other for Kotlin. In this case you should configure all the possible annotation processors, even if in fact you are not using some of them. This should not be too much problematic because if the annotation is not in your code, the processor do nothing.

Here you can see our example of this:
 - https://github.com/autentia/micronaut-base-parent
 - https://github.com/autentia/micronaut-java-parent
 - https://github.com/autentia/micronaut-kotlin-parent

graemerocher:
Yeah but I think that doesn't work if you need to customize the processors in the child

alejandropg:
Coooooorrect, the best example is if you want to use Lombok, where you need to ensure that their annotation processor is before of Micronaut.

Other way could be to modify the annotation processor plugin to be more flexible. But I don't know if it's worth it.

PR #354, opened 2020-07-27
Rename kubernetes to config-kubernetes. Fixes #353

graemerocher:
Fixes #353

PR #356, opened 2020-07-27
Split k8s yaml from kubernetes config feature. Fixes #355

### LINKED ISSUE #355: Generate basic k8s service config when selecting kubernetes feature
We should generate a basic k8s config when creating an app with the `kubernetes` feature.

graemerocher:
Fixes #355

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/Category.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/Category.java
@@ -25,6 +25,7 @@ public class Category {
     public static final String API                  = "API";
     public static final String IOT                  = "Internet of Things";
     public static final String CACHE                = "Cache";
+    public static final String CLOUD                = "Cloud";
     public static final String RESILIENCE           = "Resilience";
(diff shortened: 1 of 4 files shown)


graemerocher:
Requires #354 to be merged first

graemerocher:
Nothing, it is just out of the box applications with Micronaut + `--features kubernetes` fail with health check problems due to the endpoints being forbidden on OKE (oracle's kubernetes) which is a bad user experience.

It probably requires extra permissions and setup to get working. Config watch and distributed config support is a very specific use case and doesn't need to to the default so I think a split of `kubernetes` and `config-kubernetes` make sense.

alvarosanchez:
Agreed on the split.

Regarding the permissions, it likely needs the ones described in https://micronaut-projects.github.io/micronaut-kubernetes/latest/guide/#_kubernetes_api_authentication

graemerocher:
For some reason the checks are showing as not completed even though they are completed if you look at the `Actions` tab 🤔

PR #359, opened 2020-07-28
Fixing command path to work on Windows

cassadyj:
Changes the command issued to include the full path to mvnw.bat when running on Windows.


test-utils/src/main/groovy/io/micronaut/starter/test/CommandSpec.groovy
```diff
@@ -60,14 +60,15 @@ abstract class CommandSpec extends Specification {
 
     String executeMaven(String command) {
         if (OperatingSystem.current.isWindows()) {
-            command = "mvnw.bat " + command
+            command = dir.getAbsolutePath()+"\\"+"mvnw.bat " + command
```
alvarosanchez, line 63: Is this supposed to be working on Unix?

cassadyj: Yes, I only updated the section where OperatingSystem.current.isWindows.  I made no modifications to anything linux/Unix based.

alvarosanchez: Oh, I'm sorry, missed the part about being Windows-specific.

PR #373, opened 2020-08-04
add tini to graalvm alpine dockerfile

eisig:
The sh in alpine would not handle to signals such "CTRL+C" or "docker stop". 
And [tini](https://github.com/krallin/tini) to handle  it.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/graalvm/template/dockerfile.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/graalvm/template/dockerfile.rocker.raw
@@ -20,7 +20,8 @@ WORKDIR /home/app/@project.getName()
 RUN native-image --no-server -cp @jarFile
 
 FROM frolvlad/alpine-glibc
-RUN apk update && apk add libstdc++
+RUN apk update && apk add libstdc++ tini
 EXPOSE 8080
 COPY --from=graalvm /home/app/@project.getName()/@project.getName() /app/@project.getName()
-ENTRYPOINT ["/app/@project.getName()"]
+ENTRYPOINT ["/sbin/tini", "--"]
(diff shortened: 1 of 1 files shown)


eisig:
I don't use Alpine a lot. I first reported an issue here https://github.com/Docker-Hub-frolvlad/docker-alpine-glibc/issues/42.
As described in that issue, we can not stop it normally.

For example, you would not see "onShutdownEvent", if you add this code.
```
    @Singleton
    public static class SampleEventListener {
        private static final Logger logger = LoggerFactory.getLogger(SampleEventListener.class);

        @EventListener
        public void onShutdownEvent(ShutdownEvent event) {
            // shutdown logic here
            logger.info("onShutdownEvent");
        }
    }
```

As  I know, https://github.com/Yelp/dumb-init And  https://github.com/krallin/tini are tinny 'init'  for containers.

Another option,as described in https://github.com/krallin/tini; for docker  1.13 or greater,  use `docker run --init`, but most people do not use it. But in some situation (for example, k8s), the option '--init' can not be use.

@ilopmar , if you think it is the right solution， I will fix the tests.

ilopmar:
So, you want this to be able to stop the docker container properly without killing it, right?

GraalVM 20.1.0 added a new flag `--install-exit-handlers` (see https://medium.com/graalvm/graalvm-20-1-7ce7e89f066b) to be able to stop the container with a `CTRL+C` properly. I tried sometime ago and it worked locally but I think it failed on CI or in other place. I don't remember the specifics.
I think it's worth it to give it another try because if it works properly we will include it by default in Micronaut so users don't need to do anything.

eisig:
That seems better.

ilopmar:
I've created https://github.com/micronaut-projects/micronaut-core/issues/3880 for this.

Issue #349, opened 2020-07-20
create-function-app --features google-cloud-function README is wrong


jameskleeh:
The README information is for create-app and doesn't work for create-function-app

`./gradlew runFunction` doesn't work with raw functions

`--entry-point io.micronaut.gcp.function.http.HttpFunction` That entry point class doesn't exist. The entry point should be the default function class

PR #394, opened 2020-08-14
don't include in README instructions to run locally if it is function

sdelamo:
Close: #349


sdelamo:
Merged since tests passed

PR #397, opened 2020-08-14
refactor: del value from @Controller("/")

sdelamo:


sdelamo:
Merging since test pass

PR #411, opened 2020-08-24
refactor: %s/isKotlinTest/isKotlinTestFramework

sdelamo:
It allows for future Kotlin related test framework to be used in the method implementation of isKotlinTestFramework.

For example:

```java
default boolean isKotlinTestFramework() {
        return getTestFramework() == TestFramework.KOTLINTEST ||   getTestFramework() == TestFramework.KOTEST;
}
```

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/annotationProcessor.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/annotationProcessor.rocker.raw
@@ -6,7 +6,7 @@ String artifact
 )
 
 @if (features.language().isJava()) {
-@if (features.testFramework().isKotlinTest()) {
+@if (features.testFramework().isKotlinTestFramework()) {
     kapt(@artifact)
 } else {
     annotationProcessor(@artifact)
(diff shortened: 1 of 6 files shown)


ilopmar:
Why do we need more than one test framework for Kotlin? We are going to deprecate `Kotlin Test` in favor of `Kotest` but I don't see why we should have both of them at the same time... Can you please explain this?

sdelamo:
I don't know if we are going to remove KotlinTest  immediately.
Moreover, we may decide support another Kotlin test framework (E.g. [Spek](https://www.spekframework.org) in the future. 

This method is used in `rocker.raw` templates to apply, for example, the Kotlin Gradle plugin. 

I prefer to keep to have a generic for any Kotlin Framework. If we add a new testing kotlin framework in the future, with this change we will not need to edit the `rocker.raw` templates in multiple place but just this method.

ilopmar:
> I don't know if we are going to remove KotlinTest immediately.

We are: https://github.com/micronaut-projects/micronaut-starter/issues/261

> Moreover, we may decide support another Kotlin test framework (E.g. Spek in the future.

This would mean to add more than one test framework per language. Is like if we decide to add TestNG also for Java.

sdelamo:
Removing KotlinTest or not is independent from this PR? Are you going to approve it or should I work around this?

ilopmar (approved):

PR #413, opened 2020-08-24
feat: supported languages and default language for test framework

sdelamo:


starter-core/src/main/java/io/micronaut/starter/options/TestFramework.java
```diff
@@ -49,4 +52,38 @@ public String getSourcePath(String path, Language language) {
...
+                return Collections.singletonList(Language.GROOVY);
+            case KOTLINTEST:
+                return Collections.singletonList(Language.KOTLIN);
+            case JUNIT:
+                return Arrays.asList(Language.JAVA, Language.KOTLIN, Language.GROOVY);
+            default:
+                throw new RuntimeException("No list of supported languages have been defined for " + this.getName());
+        }
+    }
+
```
ilopmar, line 72: I already did a comment before (I can't find it anymore, maybe because of the push -f). Why do we only allow Spock for Java applications? We agreed sometime ago to support every combination of language and test framework and with this we won't allow that anymore.

Can you explain the reason behind this change?

ilopmar (changes requested):

starter-core/src/main/java/io/micronaut/starter/options/TestFramework.java
sdelamo: What this mean is the language in which the test are written for any given test framework ?

For spock we write templates only in Groovy, for Junit we generate tests in three different languages, for Kotlin test only in groovy.

sdelamo: What this mean is the language in which the test are written for any given test framework.

For spock we write templates only in Groovy, for Junit we generate tests in three different languages, for Kotlin test only in groovy.

ilopmar: I still don't understand and maybe it's because the name of the method is not the best `getSupportedLanguages`. 

> For spock we write templates only in Groovy, for Junit we generate tests in three different languages, for Kotlin test only in groovy.

We don't use Groovy or Kotlin to write JUnit tests, we only use Java...

sdelamo: Example of Junit Kotlin tests: 

Screenshot 2020-08-24 at 13 14 09

Same for Groovy and Junit.

sdelamo: >> I still don't understand and maybe it's because the name of the method is not the best getSupportedLanguages.

I am open to a better naming.

ilopmar: So, from what I remember, we decide to set JUnit as default for Kotlin in Micronaut 2.0 because Kotlin Test was deprecated in favor of Kotest and we didn't have time to do the migration. We I think we should do for Micronaut 2.1 is set Kotest as the default testing framework for Kotlin and remove/convert all those files to Kotlin-Kotest.

Regarding the name, I don't have a better one because, as I said, I don't see the usefulness of this method and where are we going to call it. But calling it `getSupportedLanguages` is miss-leading because we support Java + Groovy and the test says otherwise.

sdelamo: > So, from what I remember, we decide to set JUnit as default for Kotlin in Micronaut 2.0 because Kotlin Test was deprecated in favor of Kotest and we didn't have time to do the migration. We I think we should do for Micronaut 2.1 is set Kotest as the default testing framework for Kotlin and remove/convert all those files to Kotlin-Kotest.

I think we should move this to an issue, if you want to change that behaviour.

Now, if you select groovy, gradle plus Junit we generate the Junit tests in Groovy which I think it is the desired scenario for the user and I think we should continue to do so.

sdelamo: What I suggested is that for Guides, for Kotlin we generate the test in Junit which speeds app guide creation.

ilopmar: > Now, if you select groovy, gradle plus Junit we generate the Junit tests in Groovy which I think it is the desired scenario for the user and I think we should continue to do so.

Sorry about that, you're right.


> What I suggested is that for Guides with Kotlin as its language, we generate the test in Junit which speeds up guide creation.

Ohh, you are doing this changes with the Guide generation in mind! I think it makes sense, but it would be even faster to use Spock everywhere, and that way we also promote it to users that don't know it.

ilopmar (approved):

starter-core/src/main/java/io/micronaut/starter/options/TestFramework.java
sdelamo: >Ohh, you are doing this changes with the Guide generation in mind! I think it makes sense, but it would be even faster to use Spock everywhere, and that way we also promote it to users that don't know it.
 
I think we should check the drop down in the UI. If you check Groovy, we should select automatically spock but give the uesr the freedom to change it to Junit.

PR #439, opened 2020-09-04
Integrate the new Gradle plugin

graemerocher:
PR for the new Gradle plugin.

Some notes:

* The BOM doesn't seem to be working like we think it is working for Maven. I had to force import of the Jackson BOM otherwise Jackson versions conflict https://github.com/micronaut-projects/micronaut-starter/pull/439/files#diff-aca5dd72ca236ffc337a4144027b7d18R70
* The Dockerfiles are gone since they are redundant 
* The shadow plugin is disabled by default but enabled explicitly for Google Cloud Function and Lambda only
* We need to sort out the branch situation in `micronaut-starter`, currently there is no 2.0.x release. Also for the Launch UI we will soon need a radio button to try the 2.1 snapshot or to switch the snapshot to be 2.1.x instead of 2.0.x


graemerocher:
The one failing test seems an intermittent, unrelated failure

ilopmar:
I think we should create the same branch structure here as in core because this project is totally coupled with every new release in core.

graemerocher:
Ok this is ready once we spin off a branch for 2.0.x

alvarosanchez:
Regarding the BOM issue, are you saying that BOMs imported from the Micronaut BOM aren't imported "transitively" from the user application?

alvarosanchez:
Also, removing the Dockerfiles will block a release until the Maven plugin is ready.

graemerocher:
I need to try and replicate again, it may have been related to another issue with the GCP invoker

graemerocher:
We can alter this to restore the Docker files just for Maven, but progress needs to be made on the Maven plugin soon otherwise it will need to be pushed to 2.2

alvarosanchez:
I'm ok with this PR as is

graemerocher:
Can we merge this and later @alvarosanchez can apply the changes for Maven since they are coming?

jameskleeh:
We had a discussion the past engineering meeting about the dependencies the
plugin is automatically configuring and would like to discuss on Tuesday

graemerocher:
Can you provide a summary of the concerns?

PR #464, opened 2020-09-16
Add support for JDK15 and deprecate JDK14

ilopmar:
See https://github.com/micronaut-projects/micronaut-core/issues/4113


ilopmar:
After fixed master branch, all the test pass here and this is ready to review and merge.

PR #475, opened 2020-09-22
Oracle Function Support

graemerocher:
Follow up PR to #439

Adds Oracle Function support for HTTP and Raw functions


starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
```diff
@@ -114,35 +133,37 @@ dependencies {
...
-    @dependency.template("io.micronaut.azure", "micronaut-azure-function-http", "implementation", null)
-} else {
     @dependency.template("io.micronaut.azure", "micronaut-azure-function", "implementation", null)
-}
+    }
 }
 @if (features.contains("http-client")) {
     @dependency.template("io.micronaut", "micronaut-http-client", "implementation", null)
+} else {
+    @dependency.template("io.micronaut", "micronaut-http-client", "testImplementation", null)
```
jameskleeh, line 166: This should be moved down to the testImplementation section as we currently are ordering deps by scope

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
```diff
@@ -495,18 +513,18 @@ dependencies {
 @if (features.contains("hibernate-gorm")) {
     @dependency.template("org.apache.tomcat", "tomcat-jdbc", "runtimeOnly", null)
 }
-
-@io.micronaut.starter.feature.build.gradle.templates.testFrameworks.template(features.language(), features.testFramework())
+@if (features.testFramework().isSpock()) {
+    @if (features.contains("aws-lambda") || features.contains("aws-lambda-custom-runtime")) {
+    testImplementation("io.micronaut:micronaut-function")
```
jameskleeh, line 518: Use @dependency.template here

starter-core/src/main/java/io/micronaut/starter/feature/function/oraclefunction/OracleFunction.java
```diff
@@ -0,0 +1,162 @@
...
+    @Override
+    public String getThirdPartyDocumentation() {
+        return "https://docs.cloud.oracle.com/en-us/iaas/Content/Functions/Concepts/functionsoverview.htm";
+    }
+
+    @Override
+    protected String getRunCommand(BuildTool buildTool) {
+        if (buildTool == BuildTool.MAVEN) {
+            return "mvnw mn:run";
+        } else {
```
jameskleeh, line 104: Use `else if (buildTool == BuildTool.GRADLE)`. If/when we add more build tools this would be faulty logic

PR #495, opened 2020-10-02
Push to GitHub

pgressa:


starter-api/build.gradle
```diff
@@ -10,7 +10,9 @@ dependencies {
     api "io.micronaut:micronaut-inject"
     api "io.micronaut:micronaut-runtime"
     api "io.micronaut:micronaut-http"
+    api "io.micronaut:micronaut-http-client"
     api "io.micronaut:micronaut-validation"
+    api 'org.eclipse.jgit:org.eclipse.jgit:5.9.0.202009080501-r'
```
pgressa, line 15: Should I move this to BOM ?

graemerocher: No

starter-api/src/main/java/io/micronaut/starter/api/create/github/client/v3/GitHubApiOperations.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author Pavol Gressa
+ * @since 2.1
+ */
+public interface GitHubApiOperations {
+
+    @Post(value = "/user/repos", single = true)
+    Flowable<GitHubRepository> createRepository(
+            @Header(HttpHeaders.AUTHORIZATION) String oauthToken,
+            @Body GitHubRepository gitHubRepository);
```
graemerocher, line 33: Since you are justing calling `blockingFirst()` on the results of these methods you can just remove the reactive types since they add no value

starter-api/src/main/java/io/micronaut/starter/api/create/zip/ZipCreateController.java
```diff
@@ -13,65 +13,80 @@
...
 /**
- * Implements the {@link CreateOperation} interface for applications.
+ * Implements the {@link ZipCreateOperation} interface for applications.
  *
  * @author graemerocher
  * @since 1.0.0
  */
 @Controller
-public class CreateController extends AbstractCreateController implements CreateOperation {
+public class ZipCreateController extends AbstractCreateController implements ZipCreateOperation {
```
graemerocher, line 60: We should modify the controller to include `@ExecuteOn(TaskExecutors.IO)`

starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.List;
+
+/**
+ * GitHub create controller.
+ *
+ * @author Pavol Gressa
+ * @since 2.1
+ */
+@Controller
+public class GitHubCreateController implements GitHubCreateOperation {
```
graemerocher, line 42: We should modify the controller to include `@ExecuteOn(TaskExecutors.IO)`

starter-web-netty/README.md
```diff
@@ -33,7 +33,7 @@ $ docker push gcr.io/[PROJECT ID]/micronaut-starter
 You are now ready to deploy your application:
 
 ```
-$ gcloud run deploy --image gcr.io/[PROJECT ID]/micronaut-starter --platform=managed --allow-unauthenticated
+$ gcloud run deploy --image gcr.io/[PROJECT ID]/micronaut-starter --set-env-vars=GITHUB_OAUTH_APP_CLIENT_ID=[CLIENT ID],GITHUB_OAUTH_APP_CLIENT_SECRET=[SECRET ID] --platform=managed --allow-unauthenticated
```
graemerocher, line 36: We believe `set-env-vars` removes any existing vars as well, think we should change this to `--update-env-vars` which updates but doesn't remove.

graemerocher (changes requested):

graemerocher:
@JasonTypesCodes @jameskleeh This is the backend for "Push to Github" functionality, to finish this feature we need someone with React knowledge for the UI, can you help with that?

JasonTypesCodes:
Yes.  I'll follow up.

starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private final GitHubCreateService gitHubCreateService;
+
+    public GitHubCreateController(GitHubCreateService gitHubCreateService) {
+        this.gitHubCreateService = gitHubCreateService;
+    }
+
+    /**
+     * Creates an application, generating a ZIP file as the response.
```
eahrold, line 51: This should be JSON

starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java
```diff
@@ -0,0 +1,77 @@
...
+     * @param type     The application type The application type
+     * @param name     The name of the application The name of the application
+     * @param features The features The chosen features
+     * @param build    The build type (optional, defaults to Gradle)
+     * @param test     The test framework (optional, defaults to JUnit)
+     * @param lang     The language (optional, defaults to Java)
+     * @return A ZIP file containing the generated application.
+     */
+    @Override
+    @Get(uri = "/github/{type}/{name}{?features,lang,build,test,javaVersion,code,state}", produces = MediaType.APPLICATION_JSON)
```
eahrold, line 62: @return should be JSON

eahrold:
I saw the note regarding the "unguessable" random string, https://github.com/micronaut-projects/micronaut-starter-ui/issues/13#issuecomment-702969172

> * redirect user to: `https://github.com/login/oauth/authorize?client_id=[CLIENT-ID]&scope=user,repo&redirect_uri=https://launch.micronaut.io/github/[APPLICATION-TYPE/[APPLICATION-NAME]&[...]state=[unguessable random string]`

And I believe the starter itself would need to generate the unguessable string in order to track that during the OAuth flow. We could possibly provide that via a `/github/config` endpoint, but a better solution may be to simply have  the starter generate and provide a redirect so the button could be `<a href="https://launch.micronaut.io/github/make/[APPLICATION-TYPE/[APPLICATION-NAME]&[...]">` and the the starter would fill in the `client_id` and `state` and issue a redirect for the front-end to blindly follow.

graemerocher:
@eahrold seems like a better approach, what do you think @pgressa ?

pgressa:
@eahrold yes, that makes sense, so I'll prepare that endpoint that will directly generate the redirect. Good point thanks!

starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java
```diff
@@ -0,0 +1,80 @@
...
+            @Nullable List<String> features,
+            @Nullable BuildTool build,
+            @Nullable TestFramework test,
+            @Nullable Language lang,
+            @Nullable JdkVersion javaVersion,
+            @NonNull String code,
+            @NonNull String state,
+            @Nullable @Header(HttpHeaders.USER_AGENT) String userAgent) {
+        GitHubRepository repository = gitHubCreateService.creatApp(type, name, features, build, test, lang, javaVersion, code, state, userAgent);
+        return new GitHubCreateDTO(repository.getUrl(), repository.getCloneUrl(), repository.getHtmlUrl());
```
eahrold, line 78: @pgressa this is the one other place where we may need to return a redirect. I guess it all depends on how we set up the GH OAuth App.  

If the the OAuth app is being set to do the callback to the **starter** here this will need to be a redirect, instead of JSON.  However if the GH OAuth App is getting configured to do the redirect to to **launch** the I will be able to just parse the redirect from GH on the UI and then make a call to this ending and poll accordingly.

I'm fine either way, just want to make sure we're on the same page.

starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java
```diff
@@ -63,18 +75,42 @@ public GitHubCreateController(GitHubCreateService gitHubCreateService) {
...
+            return HttpResponse.ok(new GitHubCreateDTO(repository.getUrl(), repository.getCloneUrl(), repository.getHtmlUrl()));
+        }
+    }
+
+    private URI constructRedirectUrl(RequestInfo requestInfo) {
+        URI redirectUri = UriBuilder.of(requestInfo.getServerURL())
+                .path(requestInfo.getPath())
+                .build();
+
+        return UriBuilder.of("https://github.com/login/oauth/authorize")
```
pgressa, line 109: This should be taken ideally from service discovery

starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java
pgressa: @eahrold sure, check this commit https://github.com/micronaut-projects/micronaut-starter/pull/495/commits/1754c5164d2301e412f5082fab449bd638ec030d

eahrold: I'm not sure this is working as expected.  I've got an Oauth App Set up in Github

Screen Shot 2020-10-06 at 10 07 58 PM

Then from a browser I go to http://localhost:8080/github/default/demo?build=gradle&lang=java

at this point I get the correct redirect to GH https://github.com/login/oauth/authorize?scope=user%2Crepo&redirect_uri=http%3A%2F%2Flocalhost%3A8080%2Fgithub%2Fdefault%2Fdemo&state=d47764fb-f1d1-454a-8ef7-f0023a0dd7cc&client_id=bfd389a24260f5fed94e. (side note I think this is missing the other url parameters from the initial request such as **build** and **lang** above)

The app authorizes and the

n calls back to the `redirect_url` which successfully generates the Repo on GitHub (awesome) but the it responds to the browser with a json string. 

I think the final response [Here](https://github.com/micronaut-projects/micronaut-starter/blob/1754c5164d2301e412f5082fab449bd638ec030d/starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java#L100) also needs to be a redirect back to the `launch`.  

I've seen it where the original request origin is bound to the unique`state` for the final redirect lookup at this point. I'm not sure the best approach.

Here's a screen cast of how it's working now.  (I'd already authorized the app so it just skips over the GH auth that part of the flow.)

Redirect mov

pgressa: True, since these are independent apps, so I have to redirect the user back to launch UI. What is your preference of the redirect target so you can render the success/failure of the action?

> side note I think this is missing the other url parameters from the initial request such as build and lang above

Will write test for it just for sure.

eahrold: It could be simply to the root with a query parameter of the repo

Http://flow.origin?url=...&cloneUrl=...&htmlUrl=...

pgressa: ok will do

pgressa: @eahrold ok so now if the conditions are met, which given the circumstances are everytime at this moment, the endpoint basically returns just 307's . First is for authorisation callback and second is to launcher. Check https://github.com/micronaut-projects/micronaut-starter/pull/495/commits/f6783dbee56562c7d03800b47461b6bace8ed6ca

.github/workflows/gcr-snapshot.yml
```diff
@@ -27,4 +27,4 @@ jobs:
       - name: Deploy to Cloud Run
         run: |
           gcloud components install beta --quiet
-          gcloud beta run deploy ${{ secrets.GCLOUD_SNAPSHOT_APP_NAME }} --quiet --image gcr.io/${{ secrets.GCLOUD_PROJECT }}/${{ secrets.GCLOUD_SNAPSHOT_APP_NAME }}:snapshot --project ${{ secrets.GCLOUD_PROJECT }} --region  us-central1 --platform managed --allow-unauthenticated --service-account=${{ secrets.GCLOUD_EMAIL }}
+          gcloud run deploy ${{ secrets.GCLOUD_SNAPSHOT_APP_NAME }} --quiet --image gcr.io/${{ secrets.GCLOUD_PROJECT }}/${{ secrets.GCLOUD_SNAPSHOT_APP_NAME }}:snapshot --project ${{ secrets.GCLOUD_PROJECT }} --region  us-central1 --update-env-vars=HOSTNAME="launch.micronaut.io",GITHUB_OAUTH_APP_CLIENT_ID=${{ secrets.GH_OAUTH_APP_CLIENT_ID }},GITHUB_OAUTH_APP_CLIENT_SECRET=${{ secrets.GH_OAUTH_APP_CLIENT_SECRET }} --platform managed --allow-unauthenticated --service-account=${{ secrets.GCLOUD_EMAIL }}
```
pgressa, line 30: @graemerocher I had to insert here injection of hostname of the launcher, because the RequestInfo contained localhost:8080. I can check whether it doesn't contain X- headers so it can be taken from there, but this way, it follows what is configured in StarterConfiguration:
```
String hostname = System.getenv(Environment.HOSTNAME);
```

starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java
```diff
@@ -0,0 +1,119 @@
...
+
+import javax.validation.constraints.Pattern;
+import java.net.URI;
+import java.util.List;
+
+/**
+ * GitHub create controller.
+ *
+ * @author Pavol Gressa
+ * @since 2.1
```
pgressa, line 49: I guess it should be now 2.2

graemerocher: yup

starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java
eahrold: Happy path is working great.  But still getting JSON when the error is thrown due to duplicate repo ( which definitely happens)

https://github.com/micronaut-projects/micronaut-starter/pull/495/files#diff-c59a5ffd3f9a9658e42605a866b52763R140

We may want to throw a custom GitHubException so we can Narrowly handle and redirect with something like ‘?error=someerrormessage’

pgressa: sure will do that

.github/workflows/gcr-snapshot.yml
graemerocher: @jameskleeh could you check if the production instance has any other environment variables set, since we may need to ensure these are set by this command

starter-api/src/main/java/io/micronaut/starter/api/create/github/GitHubCreateController.java
pgressa: @eahrold error handling should be there.

eahrold: @pgressa Looks good.  Here's the ui companion.
https://github.com/micronaut-projects/micronaut-starter-ui/pull/14

graemerocher:
I am going to merge this for now, I think it has gone through enough review cycles and we need to get it tested with the snapshot

PR #507, opened 2020-10-07
Allow use any JDK version to create app from the CLI

### LINKED ISSUE #498: micronaut 2.1.0 does not work with JDK 14 but work with JDK {8,11}
Per https://github.com/Homebrew/homebrew-core/pull/62020, I built and installed mn in my local, but I could not have it run with JDK14, but it works with JDK 8 or 11 though.

```
$ jenv local 1.8

$ /usr/local/Cellar/micronaut/2.1.0/bin/mn create-app hello-world
| Application created at /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/hello-world

$ jenv local 11.0

$ /usr/local/Cellar/micronaut/2.1.0/bin/mn create-app hello-world
| Application created at /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/hello-world

$ jenv local 14.0

$ /usr/local/Cellar/micronaut/2.1.0/bin/mn create-app hello-world
| Error Cannot create the project because the target directory already exists

$ jenv versions
  system
  1.8
  1.8.0.252
  11.0
  11.0.6
* 14.0 (set by /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/.java-version)
  14.0.1
  openjdk64-1.8.0.252
  oracle64-11.0.6
  oracle64-14.0.1
```

### LINKED ISSUE #506: Support all JDK versions from CLI
I think it is confusing for the users the current behavior regarding the supported JDK versions.

It is not possible to create a new application using the CLI with the not supported JDK versions. See https://github.com/micronaut-projects/micronaut-starter/blob/078a790b7d678f3dd0d28057f62e834db1e6e017/starter-core/src/main/java/io/micronaut/starter/options/JdkVersion.java#L28-L35

If you try it, it fails with:

```
$ mn create-app foo --jdk=14
| Error Unsupported JDK version: 14. Supported values are [8, 9, 10, 11, 12, 13, 14, 15]
```
The fix for the error message is here https://github.com/micronaut-projects/micronaut-starter/pull/505

On the other hand, it is possible to create a new application with the deprecated JDK versions using Launch webpage (and it shouldn't be possible for the sake of consistency).

I think we should completely remove support of deprecated JDK versions (as Spring does) and only allow 8, 11 and 15. Once 16 is out in 6 months we completely remove 15 and add 16. With 17 being an LTS version we would add it and keep it and then start with the same approach for 18, 19,...

This change should also be done not only here but in the UI to remove those versions from the dropdown.

I don't see any valid use case for allowing JDK 13 or 14. If users want to use any of those deprecated versions for new applications, they are free to generate an application with 11 (or 15) and then manually modify the version. I think we should encourage the usage of only supported versions.

This kind of confusion already happened here: See https://github.com/micronaut-projects/micronaut-starter/issues/498

ilopmar:
Fixes #498 Fixes #506

---

I've removed the `hasSupport` boolean because we are not using it in any place and it's misleading because it is possible to create applications with "unsupported" JDKs. Besides that, if we allow to use any JDK version, there are no "unsupported" versions (as long as they exist) in the enum.

The CLI didn't allow to create apps with "unsupported versions" because the `getJdkVersion` method here
https://github.com/micronaut-projects/micronaut-starter/blob/43a3acc5be0ea532a3574efd919b9a746217def6/starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateCommand.java#L122

eventually calls the following, that only allow the so called supported JDKs:

https://github.com/micronaut-projects/micronaut-starter/blob/acc2d80c4bff32a123c92ddce91c6e9987bdb6f6/starter-core/src/main/java/io/micronaut/starter/options/JdkVersion.java#L45-L55

On the other hand, the API directly receives a `JdkVersion` as parameter

https://github.com/micronaut-projects/micronaut-starter/blob/847bb5781616a3853aa4f9df5107c1ddf99e38d7/starter-api/src/main/java/io/micronaut/starter/api/create/AbstractCreateController.java#L86

and it only checks if it is `null` or not to set a default JDK_8:

https://github.com/micronaut-projects/micronaut-starter/blob/847bb5781616a3853aa4f9df5107c1ddf99e38d7/starter-api/src/main/java/io/micronaut/starter/api/create/AbstractCreateController.java#L100

With this change both the CLI and the API will behave the same.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/options/JdkVersion.java
+++ b/starter-core/src/main/java/io/micronaut/starter/options/JdkVersion.java
@@ -16,6 +16,7 @@
 package io.micronaut.starter.options;
 
 import java.util.Arrays;
+import java.util.List;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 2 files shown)


ilopmar:
The failure is not related to this PR. I've restarted the CI twice and every time a different job has failed for a taking too much time downloading the dependencies.

PR #516, opened 2020-10-09
Only generate config in bootstrap.yml with discovery and config Consul

### LINKED ISSUE #508: Problem With Config Consul Plus Discovery Consul
### Description

If a project is created using both the `config-consul` and `discovery-consul` features, the `consul.client.registration.enabled` and `consul.client.defaultZone` config settings are defined in 2 places, causing confusion because changing the value of one of them has no affect.

### Steps to Reproduce

```
~ $ mn create-app -f discovery-consul,config-consul consuldemo
| Application created at /Users/jeffscottbrown/consuldemo
~ $ 
~ $ cat consuldemo/src/main/resources/application.yml 
consul:
  client:
    registration:
      enabled: true
    defaultZone: ${CONSUL_HOST:localhost}:${CONSUL_PORT:8500}
~ $ 
~ $ cat consuldemo/src/main/resources/bootstrap.yml 
micronaut:
  application:
    name: consuldemo
  config-client:
    enabled: true
consul:
  client:
    registration:
      enabled: true
    defaultZone: ${CONSUL_HOST:localhost}:${CONSUL_PORT:8500}
```

### Environment Information

- **Operating System**: macOs
- **Micronaut Version:** 2.1.0
- **JDK Version:** openjdk 11.0.2 2019-01-15

ilopmar:
Fixes #508

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/discovery/Consul.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/discovery/Consul.java
@@ -39,8 +39,10 @@ public class Consul implements DiscoveryFeature {
 
     @Override
     public void apply(GeneratorContext generatorContext) {
-        generatorContext.getConfiguration().put("consul.client.registration.enabled", true);
-        generatorContext.getConfiguration().put("consul.client.defaultZone", "${CONSUL_HOST:localhost}:${CONSUL_PORT:8500}");
+        if (!generatorContext.isFeaturePresent(io.micronaut.starter.feature.distributedconfig.Consul.class)) {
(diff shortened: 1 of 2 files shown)


jameskleeh:
@ilopmar This change doesn't make sense to me. I don't think the distributed config feature should be adding this line here https://github.com/micronaut-projects/micronaut-starter/blob/2.1.x/starter-core/src/main/java/io/micronaut/starter/feature/distributedconfig/Consul.java#L43. We can leave the registration.enabled flag for this feature. It seems to me that both features should be adding a generic non visible consul feature that adds the consul.client.defaultZone configuration. It would check if a `DistributedConfigFeature` is applied and either add it to bootstrap or normal config

PR #522, opened 2020-10-16
Bump to Gradle 6.7 / Gradle plugin 1.0.5

graemerocher:


ilopmar (changes requested):
You also need to update the "main" Gradle wrapper that we use for the Starter itself (in `gradle/wrapper` directory)

graemerocher:
Not sure how that has any relation to what we generate for end users projects

ilopmar:
@jameskleeh mentioned it the other day but I don't remember the exact reason.
I think it was some constraint on Gradle and that is the only way to run the tests with the same version used for the generated apps.

PR #527, opened 2020-10-23
Merge 2.1.x in master

ilopmar:
PR to check that all the test pass. Then I'll close it and manually merge so we keep track of these merges.


ilopmar:
Merged!

Issue #499, opened 2020-10-04
Add Documentation for Installation with MacPorts


JasonTypesCodes:
See: https://github.com/micronaut-projects/static-website/issues/169

When the documentation moved to starter, it seems that the MacPorts portion did not come along with it.

PR #528, opened 2020-10-23
#499 puts the MacPorts installation docs back in

willbuck:


willbuck:
Confirmed these directions work on Mac today, not sure what the CI failure was looked like an unrelated error though

graemerocher:
Do we publish to macports?

willbuck:
I am not 100% certain on it @graemerocher, but it would seem to be the case that we do; the version I installed successfully via this method was 2.1.1

Hope all is well with you!

alvarosanchez:
The Micronaut port file is kindly supported by @breun.

breun:
I indeed maintain the Portfile for Micronaut in MacPorts: https://github.com/macports/macports-ports/blob/master/java/micronaut/Portfile

JasonTypesCodes (approved):

PR #535, opened 2020-11-02
Dont render {} for empty yaml config

### LINKED ISSUE #384: Problem With "empty" application.yml
Under certain circumstances the starter creates an `application.yml` file that is empty except for opening and closing curly braces.

```
~ $ mn create-app someapp -f config-consul
| Application created at /Users/jeffscottbrown/someapp
~ $ 
~ $ cat someapp/src/main/resources/application.yml 
{
  }
~ $
```

The same thing happens in Launch when creating a project with the `config-consul` feature (probably others, but I have not verified).

Screen Shot 2020-08-12 at 11 13 59 AM

### Environment Information

- **Operating System**: macOs locally, also happens with launch.micronaut.io
- **Micronaut Version:** 2.0.1

jameskleeh:
Fixes #384

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/template/YamlTemplate.java
+++ b/starter-core/src/main/java/io/micronaut/starter/template/YamlTemplate.java
@@ -50,6 +50,8 @@ public class YamlTemplate implements Template {
         if (!config.isEmpty()) {
             Yaml yaml = new Yaml(options);
             yaml.dump(config, new OutputStreamWriter(outputStream));
+        } else {
+            outputStream.write("# Place application configuration here".getBytes());
         }
     }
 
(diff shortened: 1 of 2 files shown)


alvarosanchez:
While I agree with this PR, does it really fix #384?

Is it better an empty `application.yml`? I'm on the opinion that the file should not be generated.

jameskleeh:
@alvarosanchez I think the likelihood of the user not ending up using application.yml is low. Keeping the file there allows them to simply populate it whereas if it didn't exist they would have to know to create it and how to name the file. I think that would cause confusion for new users.

alvarosanchez:
I don't know, generating an empty seems odd to me, but I don't have a strong opinion on this.

PR #554, opened 2020-11-13
Adds correlative default values to Language

eahrold:
- On the UI when you change the Language, this guides the values on BuildTool/TestFramework (when untouched)

### DIFF
--- a/starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateCommand.java
+++ b/starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateCommand.java
@@ -33,6 +33,7 @@ import picocli.CommandLine;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.Callable;
 
 public abstract class CreateCommand extends BaseCommand implements Callable<Integer> {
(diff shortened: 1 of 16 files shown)


graemerocher:
Seems to only apply the API though? If the CLI is used the defaults would be different

starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateCommand.java
```diff
@@ -44,15 +45,15 @@
 
     @ReflectiveAccess
     @CommandLine.Option(names = {"-l", "--lang"}, paramLabel = "LANG", description = "Which language to use. Possible values: ${COMPLETION-CANDIDATES}.", completionCandidates = LanguageCandidates.class, converter = LanguageConverter.class)
-    Language lang;
+    Language lang = Language.DEFAULT_OPTION;
```
jameskleeh, line 48: We can't set a default language here because the language can be inferred based on the selected features

jameskleeh: The build features and test features have the knowledge of what is the default currently and they apply themselves based on the build / test options not being set

jameskleeh:
Closing this in favor of https://github.com/micronaut-projects/micronaut-starter/pull/563

eahrold:
> Closing this in favor of #563

Awesome. Thanks @jameskleeh

PR #555, opened 2020-11-13
Support for R2DBC in launch

graemerocher:


graemerocher:
The failures seem unrelated and are being caused by the build picking up the latest version of the Google Cloud Function stuff which doesn't support Java 8

PR #557, opened 2020-11-13
Issue 229 - add feature support for Project Lombok. 

wetted:
Issue 229 - add feature support for Project Lombok. 
- Note that lombok must precede micronaut-inject-java for annotation processing.


graemerocher:
@alvarosanchez is this correct with regards to maven?

starter-core/src/main/java/io/micronaut/starter/feature/other/ProjectLombok.java
```diff
@@ -0,0 +1,80 @@
...
+import io.micronaut.starter.options.Language;
+
+import javax.inject.Singleton;
+
+@Singleton
+public class ProjectLombok implements LanguageSpecificFeature {
+
+    private static final String LOMBOK_MAVEN_KEY = "lombok.version";
+    private static final String LOMBOK_GRADLE_KEY = "lombokVersion";
+    private static final String LOMBOK_VERSION = "1.18.16";
```
graemerocher, line 32: Please move this hard coded version to the BOM and call `VersionInfo.getDependencyVersion(String id)` instead

graemerocher (changes requested):

alvarosanchez:
@graemerocher it is targeting 2.1, so for 2.1 is correct.

graemerocher:
We shouldn't be adding new features to 2.1 at this point

alvarosanchez:
Right - @wetted can you please retarget it to 2.2.x?

starter-core/src/main/java/io/micronaut/starter/feature/other/ProjectLombok.java
jameskleeh: @graemerocher This is another thing like testcontainers. I'd like to get group consensus before we add this to the bom

alvarosanchez (changes requested):
Rejecting this until it is rebased and the version is moved to the BOM

wetted:
Closing this PR. I changed it re-target 2.2.x with a new source branch
See https://github.com/micronaut-projects/micronaut-starter/pull/562

PR #562, opened 2020-11-17
Issue 229 - add feature support for Project Lombok.

wetted:
Note that lombok must precede micronaut-inject-java for annotation processing.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
@@ -821,14 +821,26 @@ List<Property> properties
         <configuration>
           <!-- Uncomment to enable incremental compilation -->
           <!-- <useIncrementalCompilation>false</useIncrementalCompilation> -->
-          <annotationProcessorPaths combine.children="append">
 @if (features.contains("lombok")) {
+          <annotationProcessorPaths combine.self="override">
             <path>
               <!-- must precede micronaut-inject-java -->
               <groupId>org.projectlombok</groupId>
               <artifactId>lombok</artifactId>
               <version>${lombok.version}</version>
             </path>
+            <path>
+              <groupId>io.micronaut</groupId>
+              <artifactId>micronaut-inject-java</artifactId>
+              <version>${micronaut.version}</version>
+            </path>
+            <path>
+              <groupId>io.micronaut</groupId>
+              <artifactId>micronaut-validation</artifactId>
+              <version>${micronaut.version}</version>
+            </path>
+} else {
+          <annotationProcessorPaths combine.children="append">
 }
 @if (features.contains("graalvm")) {
             <path>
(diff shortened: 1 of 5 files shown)


wetted:
This PR replaces https://github.com/micronaut-projects/micronaut-starter/pull/557

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
```diff
@@ -819,6 +822,14 @@ List<Property> properties
           <!-- Uncomment to enable incremental compilation -->
           <!-- <useIncrementalCompilation>false</useIncrementalCompilation> -->
           <annotationProcessorPaths combine.children="append">
+@if (features.contains("lombok")) {
+            <path>
+              <!-- must precede micronaut-inject-java -->
```
alvarosanchez, line 827: This one is tricky. By default, now we inherit the annotation processor definitions from the parent POM, and the `combine.children="append"` attribute in line 824 will make the effective POM to contain [the parent elements](https://github.com/micronaut-projects/micronaut-core/blob/2.2.x/parent/build.gradle#L121-L130) first, then Lombok.

If Lombok needs to go first, then your feature needs to generate something like:

```xml
<annotationProcessorPaths combine.self="override">
  <!-- lombok -->
  <!-- micronaut-inject-java -->
  <!-- micronaut-validation -->
</annotationProcessorPaths>
```

starter-core/src/test/groovy/io/micronaut/starter/feature/other/ProjectLombokSpec.groovy
```diff
@@ -0,0 +1,93 @@
...
+        def context = buildGeneratorContext(
+                ['lombok'],
+                new Options(Language.JAVA, TestFramework.SPOCK, BuildTool.MAVEN),
+                ApplicationType.DEFAULT)
+        String template = pom.template(ApplicationType.DEFAULT, buildProject(),
+                context.getFeatures(), context.getBuildProperties().getProperties()).render().toString()
+
+        then:
+        template.contains("""
+    <lombok.version>1.18.16</lombok.version>
```
alvarosanchez, line 73: Where is this coming from? If it's in the BOM, then the property will be part of the BOM's POM., and inherited to children. It should not be generated in the end-user POMs.

alvarosanchez (changes requested):

starter-core/src/test/groovy/io/micronaut/starter/feature/other/ProjectLombokSpec.groovy
wetted: OK. It looks like we are adding similar properties for others like dekorate and openapi, so maybe we should look through other features to clean things up (unless I am missing some reason they need to be there).

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
wetted: @alvarosanchez Do I need to have two `annotationProcessorPaths` each with the different attributes, or am I changing the attribute on the existing one with `combine.self="override"` for all the `<path> child elements?

wetted: I think I answered my own question and updated the PR with a revised changeset.

alvarosanchez (approved):

PR #574, opened 2020-11-20
Fix auth token permission

pgressa:
With workflow features we need to request even workflow permissions.


pgressa:
Tested with workflow feature see https://github.com/pgressa/demoapplication
Screenshot 2020-11-20 at 12 13 29
Screenshot 2020-11-20 at 12 13 51

alvarosanchez (approved):

PR #589, opened 2020-12-03
Issue 275 - add Mockito feature

### LINKED ISSUE #275: Mockito Feature
I think it maybe useful to allow the user to add mockito as a feature: 

`    testImplementation "org.mockito:mockito-junit-jupiter:2.22.0"`

Maybe include mockito version in the bom so that the user don't have to define the version.

It may be useful to include it by default when you use Junit as the test framework.

wetted:
Fixes #275

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/test/Mockito.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/test/Mockito.java
@@ -39,20 +39,6 @@ public class Mockito implements Feature {
         return "Mockito test mocking framework for JUnit";
     }
 
-//    @Override
-//    public boolean shouldApply(ApplicationType applicationType, Options options, Set<Feature> selectedFeatures) {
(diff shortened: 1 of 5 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/test/Mockito.java
```diff
@@ -0,0 +1,70 @@
...
+//        if (selectedTest == null) {
+//            selectedTest = options.getLanguage().getDefaults().getTest();
+//        }
+//        return supports(applicationType) && selectedTest == TestFramework.JUNIT;
+//    }
+
+//    @Override
+//    public void apply(GeneratorContext generatorContext) {
+//
+//    }
```
alvarosanchez, line 54: Please remove

alvarosanchez (changes requested):

alvarosanchez (approved):

PR #596, opened 2020-12-10
Server mismatch

### LINKED ISSUE #453: Throw error for mismatch server features
If a non Micronaut server is chosen (Ktor for example), if any additional features are chosen that rely on the Micronaut server, an error should be thrown.

wetted:
Fixes #453 

Adding validation to ensure that features dependent on a Micronaut server are not combined with 3rd party servers that can't support them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,14 @@ You can build the CLI from source by [following these instructions](https://micr
 
 Avoid having complex logic to determine if a default feature should be applied. If, for example, a default feature should not apply if feature x is chosen, the default feature should not add logic to assert it should not apply if feature x is chosen. Instead, feature x should exclude the default feature.
 
+### Server Features and Server Dependent Features Logic
+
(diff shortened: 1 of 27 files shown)


wetted:
I did the best I could to determine which starter features are micronaut-0server dependent. I probably did not get it accurate: with some missing, and other included that shouldn't be.

Just let me know. It's pretty easy to make adjustments now.

alvarosanchez (approved):

ilopmar (changes requested):
I think this changes are a little bit "aggressive".

I'm not sure that the following dependencies do not work with Ktor. Maybe we should test them.

```
Consul, Micrometer, Multitenancy, SecuritySession, Jaeger, Zipkin, Views
```

On the other hand I think these are ok: 

```
OpenAPI, JaxRS, Management, SpringWeb.
```

Unless we are 100% sure that they don't work with Ktor I think it should be safe to not excluding them.

wetted:
As per discussion on eng meeting, I reverted `Consul`, but added all the security modules, since they depend on `@Filter`. `Micrometer, Multitenancy, Jaeger, Zipkin, Views` also depend on it so no further changes are needed.

ilopmar (approved):

PR #597, opened 2020-12-10
Fixing issue where graalvm feature can be selected with gcp, which is unsupported

### LINKED ISSUE #391: Micronaut CLI Project Generation  GCP Function Graalvm
### Description

When generating a project with the google-cloud-function feature, you can also include the graalvm feature.

`mn create-app my-app -f google-cloud-function -f graalvm`

The generated project compiles and will work just fine as a GCP HTTP Function, but the docker-build process will fail when attempting to build a native image.

The project does not contain any class with a main method, as expected for serverless functions.  If one wants to generate a graal native image they would need to create a project that does not use `google-cloud-function`.

Recommend to not allow the `google-cloud-function` feature along with the `graalvm` feature.

### Steps to Reproduce

1. `mn create-app my-app -f google-cloud-function -f graalvm`
2. `./gradlew assemble`
3. `./docker-build.sh`

### Expected Behaviour

Project should not contain the ability to create a docker container with a function project

### Actual Behaviour

Docker build fails
`Error: Main entry point class 'my.app.Application' not found.
`
### Environment Information

- **Operating System**: Windows 10
- **Micronaut Version:** 2.0.1
- **JDK Version:** 11

wetted:
Fixes #391

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/function/gcp/GoogleCloudFunctionFeatureValidator.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/function/gcp/GoogleCloudFunctionFeatureValidator.java
@@ -17,6 +17,7 @@ package io.micronaut.starter.feature.function.gcp;
 
 import io.micronaut.starter.application.ApplicationType;
 import io.micronaut.starter.feature.Feature;
+import io.micronaut.starter.feature.graalvm.GraalVM;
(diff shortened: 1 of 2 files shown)


alvarosanchez (approved):

ilopmar (changes requested):
We should add also that users can use "Google Cloud Run" to deploy the GraalVM native images as docker containers.

ilopmar (approved):

PR #598, opened 2020-12-11
Remove modifier 'public' from Java class definition because it is usu…

tobiasschaefer:
…ally omitted for JUnit 5 tests


ilopmar (approved):

ilopmar (approved):

ilopmar:
Thanks!

PR #623, opened 2021-01-15
add func tests under test-core for io.micronaut.starter.core.test.buldTool

wetted:


alvarosanchez (approved):

wetted:
@alvarosanchez this change enables the functional tests for `MavenPackageSpec` but fails for the `test maven Docker Native packaging` test case with the message "To build native images you must set the Java target byte code level to Java 11 or below" I wonder if we should move that into a different test class with an `@Requires` for JVM version (i.e. 8 or 11, but not 15)? The CI for 8 and 11 were cancelled when 15 failed, so I still don't know if they will pass otherwise.

alvarosanchez:
Using Spock's `@Requires` makes sense. Thanks for catching this one.

ilopmar (approved):

alvarosanchez (approved):

PR #635, opened 2021-01-26
Issue 631 - add retry to spock tests that fail intermittently due to port binding race conditions on CI

### LINKED ISSUE #631: Cloud tests intermittent failure
https://github.com/micronaut-projects/micronaut-starter/pull/630/checks?check_run_id=1749719222#step:5:229

wetted:
Fixes #631

### DIFF
--- a/test-core/src/test/groovy/io/micronaut/starter/core/test/aws/CreateAwsAlexaSpec.groovy
+++ b/test-core/src/test/groovy/io/micronaut/starter/core/test/aws/CreateAwsAlexaSpec.groovy
@@ -9,6 +9,7 @@ import io.micronaut.starter.options.TestFramework
 import spock.lang.Retry
 import spock.lang.Unroll
 
+@Retry // can fail on CI due to port binding race condition, so retry
 class CreateAwsAlexaSpec extends CommandSpec {
 
     @Override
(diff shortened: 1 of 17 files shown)


alvarosanchez:
Did you add the annotation to all the tests that bind a port? Or just the ones you have seen failing recently?

If the latter, it's maybe a better idea to add the annotation to all the spec classes, rather than test by test.

wetted:
> Did you add the annotation to all the tests that bind a port? Or just the ones you have seen failing recently?
> 
> If the latter, it's maybe a better idea to add the annotation to all the spec classes, rather than test by test.

I annotated the cloud tests. Each of those spec classes has just a single data-driven test. I guess it doesn't make a difference in this case, but I can change them.

alvarosanchez:
I just would like to get rid of this issue, and not find it weeks later that other tests are failing because of the same issue.

Another thing we should consider is to remove the `--parallel` flag. I suspect it is the source of those issues.

wetted:
Another problem I'm seeing is with integration tests that generate a project, and build the project to run the test task. Sometimes we get a connection failure/reset with Maven Central, so I'm going to add Retry to the ones I know about so far.

PR #641, opened 2021-02-02
Added Pebble support

ecmel:


starter-core/src/main/java/io/micronaut/starter/feature/view/Pebble.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2020 original authors
```
burtbeckwith, line 2: should be 2021

ecmel: Fixed

alvarosanchez (changes requested):
Thanks for the PR! A couple of things to be changed:

* Please remove `.project` and `.settings`, and add them to `.gitignore` to avoid being committed again.
* Apart from the feature class, you also need to change the build templates to include the dependency if the feature is selected:
  * Gradle: https://github.com/micronaut-projects/micronaut-starter/blob/2.3.x/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/dependencies.rocker.raw
  * Maven: https://github.com/micronaut-projects/micronaut-starter/blob/2.3.x/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw

ecmel:
@alvarosanchez Updated

alvarosanchez (approved):

ilopmar (changes requested):
You need to add tests. Please take a look at this https://github.com/micronaut-projects/micronaut-starter/blob/5928195a1f0c6abc53766851a113be01979efa56/starter-core/src/test/groovy/io/micronaut/starter/feature/view/HandlebarsSpec.groovy for handlebars

ecmel:
@ilopmar Added test

ecmel:
Pebble tests are passing

alvarosanchez:
I reran the jobs and they are all passing now.

Thank you very much for your contribution!

PR #643, opened 2021-02-04
feat: allow features to contribute dependencies

sdelamo:
This change allows a feature to contribute dependencies to both maven and grade files.  This could be used in libraries using the `starter-core` (e.g. guides). 

coordinates will be defined in a single configuration file so that dependency version numbers are not scattered in the code.


starter-core/src/main/java/io/micronaut/starter/feature/dependencies/GradleConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.starter.feature.dependencies;
+
+public enum GradleConfiguration {
```
alvarosanchez: Using an enum would prevent in the future using custom configurations (eg `developmentOnly`, `documentation`, etc).

Also, `annotationProcessor` is missing.

starter-core/src/main/java/io/micronaut/starter/feature/dependencies/ScopedMavenCoordinate.java
```diff
@@ -0,0 +1,121 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.starter.feature.dependencies;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import io.micronaut.core.annotation.Introspected;
+
+@Introspected
+public class ScopedMavenCoordinate implements MavenCoordinate {
```
alvarosanchez: All the Maven coordinates have always an scope, so I don't get why the complexity of a `MavenCoordinate` interface + this class. This could be simplified IMHO

starter-core/src/main/java/io/micronaut/starter/feature/dependencies/DependenciesFeature.java
```diff
@@ -0,0 +1,54 @@
...
+import io.micronaut.starter.feature.test.TestFeature;
+import io.micronaut.starter.options.BuildTool;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public interface DependenciesFeature {
```
alvarosanchez: Why adding these methods here? Could they not belong to `Features`?

alvarosanchez (changes requested):
Sergio, this is AWESOME. It will greatly improve our features, and reduce the amount of hardcoded versions. Thanks for putting this together.

That said, I would propose to store the dependencies coordinates and versions in a format that Dependabot can pick it up and send upgrade PRs to. For example, we could have a _naked_ (with only `<dependencies>`) POM at `starter-core/src/main/resources`.

starter-core/src/main/java/io/micronaut/starter/feature/dependencies/DependenciesConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+package io.micronaut.starter.feature.dependencies;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import io.micronaut.context.annotation.EachProperty;
+import io.micronaut.context.annotation.Parameter;
+
+@EachProperty("dependencies")
+public class DependenciesConfiguration implements MavenCoordinate {
```
ilopmar: Instead of having this class that implements the interface you could annotate the interface directly. That works for `@ConfigurationProperties` so I think it should work with `@EachProperty`

ilopmar (changes requested):

starter-core/src/test/resources/application-test.yml
```diff
@@ -0,0 +1,21 @@
+dependencies:
```
sdelamo: If understand correctly @alvarosanchez what you are suggesting is to replace the yaml configuration with a pom.xml file such as:

```xml
<project>
  <dependencies>
    <dependency>
      <groupId>org.mybatis</groupId>
      <artifactId>mybatis</artifactId>
      <version>3.4.6</version>
    </dependency>
    <dependency>
      <groupId>org.gebish</groupId>
      <artifactId>geb-junit5</artifactId>
      <version>4.0</version>
    </dependency>
    <dependency>
      <groupId>org.seleniumhq.selenium</groupId>
      <artifactId>selenium-firefox-driver</artifactId>
      <version>3.141.59</version>
    </dependency>
    <dependency>
      <groupId>org.seleniumhq.selenium</groupId>
      <artifactId>selenium-support</artifactId>
      <version>3.141.59</version>
    </dependency>
  </dependencies>
</project>
```
is that correct?

The benefit will be that dependabot will create prs automatically to such a file, right?

We will need to support multiple pom files. Starter core could have its own pom and the guides project will have an additional pom with dependencies such as `mybatis`, `sendgrid`, `spreadsheetbuilder` etc.

sdelamo (commented):
comment about pom like files

starter-core/src/test/resources/application-test.yml
alvarosanchez: Correct, that is what I suggest.

Then we could point Dependabot to that folder: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#directory

sdelamo: does `dependabot` need the file to be named `pom.xml`. ? 

should we place them in the `META-INF`? 

For example in starter-core: 

`starter-core/src/main/resources/META-INF/starter-dependencies/core/pom.xml`

and in guides: 

`buildSrc/src/main/resources/META-INF/starter-dependencies/guides/pom.xml`

I could have a factory which reads every pom.xml in the `META-INF/starter-dependency` and creates `MavenCoordinate` beans.

alvarosanchez: Yes, I think it needs to be named `pom.xml`. The directory can be whatever you want, though

sdelamo:
I am working on the changes we discussed. I think it is looking promising.

starter-core/src/main/java/io/micronaut/starter/feature/dependencies/DependenciesConfiguration.java
sdelamo: This no longer applies

starter-core/src/test/resources/application-test.yml
sdelamo: I've used a pom.xml

starter-core/src/main/java/io/micronaut/starter/feature/dependencies/GradleConfiguration.java
sdelamo: > Also, annotationProcessor is missing.

I've added annotationProcessor

starter-api/build.gradle
```diff
@@ -1,5 +1,7 @@
 dependencies {
     annotationProcessor "io.micronaut.configuration:micronaut-openapi:1.5.3"
+    annotationProcessor "com.google.code.findbugs:jsr305"
```
sdelamo: this is needed to pull javax Nullable

sdelamo:
See Log4j2 for an example of a dependency version defined in src/main/resources/pom.xml

also, consumers of the library can apply dependencies which modify the build. See MyBatis Feature.

starter-core/src/main/java/io/micronaut/starter/feature/dependencies/DependenciesFeature.java
sdelamo: This no longer applies

alvarosanchez (approved):

ilopmar (approved):

sdelamo:
Close via https://github.com/micronaut-projects/micronaut-starter/pull/679

Issue #346, opened 2020-07-16
Display binary files on the preview pane but do not allow to select them


ilopmar:
It's confusing for users not to see all the files in the preview tree (see https://github.com/micronaut-projects/micronaut-starter/issues/344).

It would be good to display all the files but do not allow to select the binary files.

For example, in the preview the `gradle-wrapper.jar` file is not displayed:

![image](https://user-images.githubusercontent.com/559192/87690707-129e4d00-c78a-11ea-9ed3-1623597bd6e8.png)

But it is included in the generated project:

![image](https://user-images.githubusercontent.com/559192/87690760-234ec300-c78a-11ea-8e30-83d3a969a644.png)

jameskleeh:
They aren't displayed because Jackson doesn't render map keys with null values

ilopmar:
@jameskleeh But we can configure Jackson for that POJO/Field to render null values. If that makes the UI display the file without the content that would be the behavior we want.

jameskleeh:
@ilopmar Right - I was just letting whoever picked this up know that is the reason why they aren't being shown now

ilopmar:
Ok. Then I'm reopening the issue.

ilopmar:
:facepalm:

PR #649, opened 2021-02-09
Display binary files on the preview pane but do not allow to select them

eahrold:
- Addresses  https://github.com/micronaut-projects/micronaut-starter/issues/346


jameskleeh:
@eahrold Does this need merged with any PR for the client?

eahrold:
> @eahrold Does this need merged with any PR for the client?

Nope, the current frontend is already coded to handle these new null values. It will show in the fs view, but not be clickable.

PR #650, opened 2021-02-10
Use Micronaut's Nullability annotations

sdelamo:
I had to remove the usage of javax.annotation.concurrent.Immutable annotation as well.


graemerocher:
This should target a 2.4.x branch I believe

PR #670, opened 2021-02-24
Update mn-windows-snapshot.yml

pgressa:


graemerocher:
looks like it is failing with a wrong java version problem

sdelamo:
This is really an old PR. I am closing it. Not sure if it still applies. @pgressa please reopen if we still need it.

PR #675, opened 2021-03-03
test: BuildBuilder to decouple test from rocker

sdelamo:
Almost every test was coupled to the rocker templates for both maven and Gradle. Modifying the arguments to those rocker files meant changing every test. I need to modify those rocker files in a future PR. 

The goal of this PR is to encapsulate the logic to call those templates in a single file `BuildBuilder` and also create nice fluid API so that tests can verify different scenarios in an easy way. 

Moreover, this PR adds:
 
- missing test that for Groovy language the Groovy Gradle plugin is applied
- missing test that for Spock even if Java or Kotlin  the Groovy Gradle plugin is applied
- some verifications that the Gradle plugin version resolved is a semantic version
- test for grpc gradle plugin
- missing test for Log4j2
- missing test for Shade plugin
- missing test for jrebel gradle plugin

I am passing the BuildBuilder constructor the application context because I will use it in a future PR. It is not used in this PR but it will in the next PR.

### DIFF
--- a/starter-core/src/test/groovy/io/micronaut/starter/ApplicationContextSpec.groovy
+++ b/starter-core/src/test/groovy/io/micronaut/starter/ApplicationContextSpec.groovy
@@ -1,0 +1,59 @@
+package io.micronaut.starter
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.core.version.SemanticVersion
+import io.micronaut.starter.fixture.ContextFixture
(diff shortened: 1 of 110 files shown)


jameskleeh:
How will the context be used in the future?

ilopmar (approved):

sdelamo:
> How will the context be used in the future?

it is shown here: https://github.com/micronaut-projects/micronaut-starter/pull/679

PR #690, opened 2021-03-09
Add coherence features

pgressa:


graemerocher (approved):

graemerocher (changes requested):
needs to override `isPreview()` for each feature and return `true`

PR #695, opened 2021-03-10
SelectOptions unit test to check JUnit label is correct.

eahrold:
- Related to https://github.com/micronaut-projects/micronaut-starter-ui/issues/42


eahrold:
@eahrold add the .yml to the resources config

https://github.com/micronaut-projects/micronaut-starter/blob/2.4.x/starter-api/src/main/resources/META-INF/native-image/io.micronaut.starter.api/micronaut-starter-api/resource-config.json

PR #704, opened 2021-03-17
feat: adds camunda feature

### LINKED ISSUE #698: Add Camunda feature
see: https://github.com/NovatecConsulting/micronaut-camunda-bpm

cc: @tobiasschaefer

sdelamo:
Close: https://github.com/micronaut-projects/micronaut-starter/issues/698

cc @tobiasschaefer

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/camunda/Camunda.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/camunda/Camunda.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.starter.feature.camunda;
+
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.starter.application.ApplicationType;
+import io.micronaut.starter.application.generator.GeneratorContext;
+import io.micronaut.starter.build.dependencies.Dependency;
+import io.micronaut.starter.feature.Category;
+import io.micronaut.starter.feature.Feature;
+import io.micronaut.starter.feature.FeatureContext;
+import io.micronaut.starter.feature.database.DatabaseDriverFeature;
+import io.micronaut.starter.feature.other.HibernateValidator;
+import io.micronaut.starter.feature.test.AssertJ;
+
+import javax.inject.Singleton;
+
+@Singleton
+public class Camunda implements Feature {
+
+    private final AssertJ assertJ;
+    private final DatabaseDriverFeature defaultDbFeature;
+    public Camunda(DatabaseDriverFeature defaultDbFeature, AssertJ assertJ) {
+        this.defaultDbFeature = defaultDbFeature;
+        this.assertJ = assertJ;
+    }
+
+    @NonNull
+    @Override
+    public String getName() {
+        return "camunda";
+    }
+
+    @Override
+    public String getTitle() {
+        return "Camunda (Workflow Engine)";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Allows to integrate Camunda (Workflow Engine) in your Micronaut applications.";
+    }
+
+    @Override
+    public boolean supports(ApplicationType applicationType) {
+        return applicationType == ApplicationType.DEFAULT;
+    }
+
+    @Override
+    public void processSelectedFeatures(FeatureContext featureContext) {
+        if (!featureContext.isPresent(DatabaseDriverFeature.class)) {
+            featureContext.addFeature(defaultDbFeature);
+        }
+        if (!featureContext.isPresent(AssertJ.class)) {
+            featureContext.addFeature(assertJ);
+        }
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.addDependency(Dependency.builder()
+                .lookupArtifactId("micronaut-camunda-bpm-feature")
+                .compile());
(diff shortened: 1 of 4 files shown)


starter-core/src/test/groovy/io/micronaut/starter/feature/camunda/CamundaSpec.groovy
```diff
@@ -0,0 +1,59 @@
...
+import io.micronaut.starter.ApplicationContextSpec
+import io.micronaut.starter.BuildBuilder
+import io.micronaut.starter.fixture.CommandOutputFixture
+import io.micronaut.starter.options.BuildTool
+import io.micronaut.starter.options.Language
+import spock.lang.Unroll
+
+class CamundaSpec extends ApplicationContextSpec implements CommandOutputFixture {
+    @Unroll
+    void 'test gradle aws-parameter-store feature for language=#language'() {
```
tobiasschaefer: "aws-parameter-store" should be "micronaut-camunda"

starter-core/src/main/java/io/micronaut/starter/feature/camunda/Camunda.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public boolean supports(ApplicationType applicationType) {
+        return applicationType == ApplicationType.DEFAULT;
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.addDependency(Dependency.builder()
+                .lookupArtifactId("micronaut-camunda-bpm-feature")
```
tobiasschaefer, line 76: when using "micronaut-camunda-bpm-feature" you also need a database. The default values use a h2. Can we please add h2 to the runtime scope automatically when selecting micronaut-camunda-bpm-feature. This would be similar to "data-jdbc".

The tests should also test that h2 is always included.

starter-core/src/main/java/io/micronaut/starter/feature/camunda/Camunda.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2020 original authors
```
tobiasschaefer, line 2: I think this should be 2021

starter-core/src/test/groovy/io/micronaut/starter/feature/camunda/CamundaSpec.groovy
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2020 original authors
```
tobiasschaefer, line 2: probably 2021

tobiasschaefer (changes requested):
Thanks @sdelamo for creating this PR. I added some comments.

starter-core/src/test/groovy/io/micronaut/starter/feature/camunda/CamundaSpec.groovy
sdelamo: done

starter-core/src/main/java/io/micronaut/starter/feature/camunda/Camunda.java
sdelamo: I have implemented so that the default database is included when you add camunda.

starter-core/src/main/java/io/micronaut/starter/feature/camunda/Camunda.java
sdelamo: We use currently 2020 all through the project. This will need to be addressed in a separate PR.

starter-core/src/test/groovy/io/micronaut/starter/feature/camunda/CamundaSpec.groovy
sdelamo: We use currently 2020 all through the project. This will need to be addressed in a separate PR.

sdelamo:
@tobiasschaefer what is the minimal configuration which we should include by default?

tobiasschaefer:
@sdelamo 
> what is the minimal configuration which we should include by default?

I think you're asking about configuration properties in application.yml, right? We actually don't require any at all and I wouldn't add any by default. 

Please update to 0.21.0 (released today).

sdelamo:
> > what is the minimal configuration which we should include by default?
> 
> I think you're asking about configuration properties in application.yml, right? We actually don't require any at all and I wouldn't add any by default.

Yes. I am talking about `application.yml`

sdelamo:
@tobiasschaefer can you clone the branch and generate a sample app locally with the feature and see if it looks good. See [Contributing running the CLI](https://github.com/micronaut-projects/micronaut-starter/blob/2.4.x/CONTRIBUTING.md#running-the-cli)

tobiasschaefer:
@sdelamo , thanks, looks quite good. Just one issue, see below.

I tested the following
* `./gradlew micronaut-cli:run --args="create-app --list-features"` --> feature is listed (OK)
* `./gradlew micronaut-cli:run --args="create-app --features=camunda temp2"` -> micronaut-camunda-bpm-feature, h2, camunda-bpm-assert and assertj are included (OK)
* `./gradlew micronaut-cli:run --args="create-app --features=camunda,postgres temp4" ` --> postgres replaces h2 (OK)
* `./gradlew micronaut-cli:run --args="create-app --build=maven --features=camunda tempm1` --> project looks good and starts (OK)

However, the order of the dependencies is a bit strange when using Gradle:

```groovy
dependencies {
    implementation("info.novatec:micronaut-camunda-bpm-feature:0.21.0")
    implementation("io.micronaut:micronaut-http-client")
    implementation("io.micronaut:micronaut-runtime")
    implementation("javax.annotation:javax.annotation-api")
    testImplementation("org.camunda.bpm.assert:camunda-bpm-assert:8.0.0")
    implementation("io.micronaut:micronaut-validation")
    runtimeOnly("ch.qos.logback:logback-classic")
    runtimeOnly("com.h2database:h2")
    testImplementation("org.assertj:assertj-core")
}
```

It would be nice if "org.camunda.bpm.assert:camunda-bpm-assert:8.0.0" was placed together with "assertj-core" which is also a "testImplementation". Can you influence that?

sdelamo:
The order will be defined by the Gradle configuration. 

implementation A-Z, testImplementation A-Z 

it is defined here:

https://github.com/micronaut-projects/micronaut-starter/blob/d14dff52b91a10b0dd132f3889b97392bd347117/starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleDependency.java#L29

We are refactoring the way dependencies are being pulled. Not every feature uses the new capabilities. That it is to say, I think ordering will improve in the short term.

PR #731, opened 2021-04-09
Add HttpClient in compile scope for aws-custom-runtime

### LINKED ISSUE #721: For Maven Applications with Custom Runtime (e.g. GraalVM)  and AWS Lambda HTTP Client should be compile not test
Currently if you generate a MIcronaut Application with features `graalvm` and `aws-lambda` and build Maven and you package with: 

`./mvnw package -Dpackaging=docker-native -Dmicronaut.runtime=lambda -Pgraalvm` 

If you deploy the generated zip file to AWS Lambda and you execute you will see that the HTTP Client class is missing:

```
START RequestId: 9e615ea2-7448-44f9-97ae-0d6e54e471ae Version: $LATEST
[36m08:06:23.958[0;39m [1;30m[main][0;39m [34mINFO [0;39m [35mi.m.f.a.p.AbstractLambdaContainerHandler[0;39m - Starting Lambda Container Handler
[36m08:06:23.958[0;39m [1;30m[main][0;39m [34mINFO [0;39m [35mi.m.context.env.DefaultEnvironment[0;39m - Established active environments: [ec2, cloud, lambda, function]
java.lang.NoClassDefFoundError: io.micronaut.http.client.DefaultHttpClientConfiguration
	at io.micronaut.function.aws.runtime.AbstractMicronautLambdaRuntime.startRuntimeApiEventLoop(AbstractMicronautLambdaRuntime.java:319)
	at io.micronaut.function.aws.runtime.AbstractMicronautLambdaRuntime.run(AbstractMicronautLambdaRuntime.java:129)
	at io.micronaut.function.aws.runtime.MicronautLambdaRuntime.main(MicronautLambdaRuntime.java:50)
Request loop failed with: io.micronaut.http.client.DefaultHttpClientConfiguration
END RequestId: 9e615ea2-7448-44f9-97ae-0d6e54e471ae
REPORT RequestId: 9e615ea2-7448-44f9-97ae-0d6e54e471ae	Duration: 326.56 ms	Billed Duration: 327 ms	Memory Size: 128 MB	Max Memory Used: 13 MB	
RequestId: 9e615ea2-7448-44f9-97ae-0d6e54e471ae Error: Runtime exited without providing a reason
Runtime.ExitError
```

For custom runtime we need the Micronaut HTTP Client to poll. See:

https://github.com/micronaut-projects/micronaut-aws/blob/master/function-aws-custom-runtime/src/main/java/io/micronaut/function/aws/runtime/AbstractMicronautLambdaRuntime.java#L330

The solution is  to change the scope of `micronaut-http-client` from `test` to `compile`

ilopmar:
Fixes #721

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/awslambdacustomruntime/AwsLambdaCustomRuntime.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/awslambdacustomruntime/AwsLambdaCustomRuntime.java
@@ -33,6 +33,7 @@ import io.micronaut.starter.feature.function.Cloud;
 import io.micronaut.starter.feature.function.CloudFeature;
 import io.micronaut.starter.feature.function.FunctionFeature;
 import io.micronaut.starter.feature.function.awslambda.AwsLambda;
+import io.micronaut.starter.feature.other.HttpClient;
 import io.micronaut.starter.template.RockerTemplate;
 
 import javax.inject.Provider;
@@ -45,9 +46,11 @@ public class AwsLambdaCustomRuntime implements FunctionFeature, ApplicationFeatu
     public static final String FEATURE_NAME_AWS_LAMBDA_CUSTOM_RUNTIME = "aws-lambda-custom-runtime";
 
     private final Provider<AwsLambda> awsLambda;
+    private final HttpClient httpClient;
 
-    public AwsLambdaCustomRuntime(Provider<AwsLambda> awsLambda) {
+    public AwsLambdaCustomRuntime(Provider<AwsLambda> awsLambda, HttpClient httpClient) {
         this.awsLambda = awsLambda;
+        this.httpClient = httpClient;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


sdelamo (commented):
This looks good. However, I am not sure why it is necessary. 

Http client should be pulled as a transitive dependency. 

https://github.com/micronaut-projects/micronaut-aws/blob/master/function-aws-custom-runtime/build.gradle#L5

ilopmar:
The problem is this: https://github.com/micronaut-projects/micronaut-starter/blob/9c7fb1327b6095cc3d841d958b8ca9e964a79220/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw#L398-L400

Unless we explictly add `http-client`, it is added in `test` scope.

sdelamo:
What I mean is, this only seems to affect Maven builds. Gradle builds resolve the http client as a transitive of `aws-custom-runtime`. Should not maven pull the http client as a transitive dependency for the compile classpath as well.

ilopmar:
Maybe what we need to do is removing that `test` dependency so the transitive `compile` is used. But in any case that won't fix the issue because if users add the `test` scoped dependency for their tests we would be in the same place like we are now.

sdelamo:
Can you try removing the test dependency in the maven build and check if the native image works ?

ilopmar:
> Can you try removing the test dependency in the maven build and check if the native image works ?

Yes, that works. The problem is that if the user adds manually the `micronaut-http-client` dependency in the `test` scope, it will fail again. That's why adding it to the `compile` scope seems the proper fix to this.

sdelamo (approved):

PR #738, opened 2021-04-16
Bump micronaut-camunda-bpm-feature to version 0.23.0 - together with …

tobiasschaefer:
…matching camunda-bpm-assert

Please bump the version of the micronaut-camunda-bpm-feature.

I'm the author of the integration, see https://github.com/NovatecConsulting/micronaut-camunda-bpm

### DIFF
--- a/starter-core/src/main/resources/pom.xml
+++ b/starter-core/src/main/resources/pom.xml
@@ -15,12 +15,12 @@
         <dependency>
             <groupId>org.camunda.bpm.assert</groupId>
             <artifactId>camunda-bpm-assert</artifactId>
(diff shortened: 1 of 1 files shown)


sdelamo:
Thanks for the PR @tobiasschaefer However, you don't need to create this PRs. In theory dependabot should create a PR automatically.

tobiasschaefer:
> Thanks for the PR @tobiasschaefer However, you don't need to create this PRs. In theory dependabot should create a PR automatically.

Thanks @sdelamo

I created the PR because both dependencies should preferably be updated together (and until now dependabot hasn't created any PRs to update those two dependences)

PR #759, opened 2021-05-03
Enable functional tests for JDK 16

graemerocher:
This enables functional tests for JDK 16, however is failing right now since Kotlin is not compatible and all the Kotlin builds are failing.


sdelamo:
We are running already for 17. https://github.com/micronaut-projects/micronaut-starter/blob/3.3.x/.github/workflows/gradle-func.yml#L14 

Closing this PR as it is out of date.

PR #777, opened 2021-05-09
feat: settings extensions

sdelamo:
This enables Features to contribute extensions to `settings.gradle`


ilopmar (approved):

jameskleeh:
@sdelamo Is there an equivalent Maven plugin?

jameskleeh:
Having a feature be specific to a build tool is something we should avoid. I think in this case someone using maven could select this feature and it just wouldn't do anything for them.

Also this check is redundant https://github.com/micronaut-projects/micronaut-starter/pull/777/files#diff-b5778c1f08a5c0e6d8ca5f784c83d3f35af41489aca36069737ad9403243d228R53

PR #801, opened 2021-05-20
MavePlugin implement equals & hashcode

sdelamo:


starter-core/src/main/java/io/micronaut/starter/feature/asciidoctor/Asciidoctor.java
```diff
@@ -59,11 +67,18 @@ public void apply(GeneratorContext generatorContext) {
...
-            generatorContext.getBuildProperties().put("jruby.version", "9.2.11.1");
+            String mavenPluginArtifactId = "asciidoctor-maven-plugin";
+            BuildProperties props = generatorContext.getBuildProperties();
+            coordinateResolver.resolve(mavenPluginArtifactId)
+                    .ifPresent(coordinate -> props.put("azure.functions.maven.plugin.version", coordinate.getVersion()));
+
+            props.put("asciidoctor.maven.plugin.version", "2.0.0-RC.1");
+            props.put("asciidoctorj.version", "2.2.0");
+            props.put("asciidoctorj.diagram.version", "2.0.1");
+            props.put("jruby.version", "9.2.11.1");
```
ilopmar: Is there a reason why we can't revolve these versions from pom.xml?

sdelamo: I've deleted `props.put("asciidoctor.maven.plugin.version", "2.0.0-RC.1");` which I had already moved to `pom.xml`.

 The other versions numbers, I don't know what to do. I don't think they correspond to maven coordinates. Moreover, they were not added as part of this PR.

ilopmar: It seems they are used with Maven:

image

As you are modifying this file, I think is worth it to move them to pom.xml

sdelamo: I have moved those dependencies to `pom.xml`

ilopmar (approved):

PR #804, opened 2021-05-21
fix: update rocker templates for azure raw function

sdelamo:
It did not work without `@FunctionName`not sure if it was always broken or we have changed anything. Do you know @graemerocher  ?

I've generated and run apps with three languages : 

```
./gradlew micronaut-cli:run --args="create-function-app --lang=java --build=gradle --features=azure-function temp"
./gradlew micronaut-cli:run --args="create-function-app --lang=groovy --build=gradle --features=azure-function temp"
./gradlew micronaut-cli:run --args="create-function-app --lang=kotlin --build=gradle --features=azure-function temp"
```

Then, I started the app:

```
cd starter-cli/temp
./gradlew azureFunctionsRun 
```

and hit the endpoint:

```
 curl localhost:7071/api/foo
Hello World
```

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/function/azure/AzureRawFunction.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/function/azure/AzureRawFunction.java
@@ -22,11 +22,14 @@ import io.micronaut.starter.application.generator.GeneratorContext;
 import io.micronaut.starter.build.dependencies.CoordinateResolver;
 import io.micronaut.starter.feature.FeatureContext;
 import io.micronaut.starter.feature.function.azure.template.azureFunctionReadme;
+import io.micronaut.starter.feature.function.azure.template.raw.azureRawFunctionHttpRequestJava;
+import io.micronaut.starter.feature.function.azure.template.raw.azureRawFunctionResponseBuilderJava;
 import io.micronaut.starter.feature.other.ShadePlugin;
 import io.micronaut.starter.options.BuildTool;
 import io.micronaut.starter.options.DefaultTestRockerModelProvider;
+import io.micronaut.starter.options.Language;
 import io.micronaut.starter.options.TestRockerModelProvider;
-
+import io.micronaut.starter.template.RockerTemplate;
 import javax.inject.Singleton;
 import java.util.Optional;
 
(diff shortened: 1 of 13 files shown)


graemerocher:
If it works 🤷‍♂️

sdelamo:
The azure docs seems to show that such a method is a HttpTrigger is valid 

```java
public class Function {
    public String echo(@HttpTrigger(name = "req", 
      methods = {HttpMethod.POST},  authLevel = AuthorizationLevel.ANONYMOUS) 
        String req, ExecutionContext context) {
        return String.format(req);
    }
}
```

https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-java?tabs=bash%2Cconsumption#triggers-and-annotations

However, it seems the signature is not correct: 

https://github.com/Azure/azure-functions-java-worker/issues/255#issuecomment-454081060

> Function signature does not look right. A HttpTrigger function should like this:

```java
public static String HttpTriggerJavaMetadata(
        @HttpTrigger(name = "req", methods = {HttpMethod.GET, HttpMethod.POST}, authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
        @BindingName("firstName") String queryValue1, @BindingName("lastName") String queryValue2
    )
```

> Note: Using @HttpTrigger annotation requires the input to be of type HttpRequestMessage which is defined in azure-functions-java-library. To specify a HttpRequest with input body as Bytes, use HttpRequestMessage<Optional<byte[]>>

Because of that I have changed the raw function  we generate to:

```java
public class Function extends AzureFunction {
    @FunctionName("HttpTrigger")
    public HttpResponseMessage hello(
            @HttpTrigger(
                    name = "req",
                    methods = {HttpMethod.GET},
                    route = "{*route}",
                    authLevel = AuthorizationLevel.ANONYMOUS)
                    HttpRequestMessage<Optional<String>> request,
            ExecutionContext context) {
        if (context != null) {
            context.getLogger().info("Executing Function: " + getClass().getName());
        }
        return request.createResponseBuilder(HttpStatus.OK).body("Hello World").build();
    }
}
```

Because microsoft azure java library does not ship with implementations of `HttpRequestMessage` I had a created implementations in Java for the test classpath to ship with a test.  I had to exclude it for Maven Kotlin. 

Ideally, they should be in Groovy / Kotlin and Java rocker templates or even better ship a 'micronaut-azure-http-test-utils' dependency with those helper classes. Nonetheless; I think we should merge this because the project we currently generate is broken.

ilopmar (approved):

PR #818, opened 2021-05-28
Fix CLI native image generation

ilopmar:
Fix CLI native image generation: https://github.com/micronaut-projects/micronaut-starter/runs/2688761956?check_suite_focus=true#step:7:152

When https://github.com/micronaut-projects/micronaut-core/pull/5503 gets merged we can remove the dependency on jakarta.inject that this PR adds and use `jakarta.inject.Provider` in Starter.

### DIFF
--- a/.github/workflows/mn-linux-snapshot.yml
+++ b/.github/workflows/mn-linux-snapshot.yml
@@ -9,7 +9,7 @@ jobs:
         uses: actions/checkout@v2.3.4
       - name: Install GraalVM
         env:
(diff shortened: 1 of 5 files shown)


ilopmar:
I've also updated the workflows to use GraalVM for JDK11 [because](https://github.com/graalvm/graalvm-ce-builds/releases/tag/vm-21.1.0):

> Note that due to the decommissioning of the aging macOS 10.7 build infrastructure, GraalVM Community Edition releases based on JDK 8 are no longer being built.

Issue #849, opened 2021-07-13
`mn create-acme-account` help output misleading about existing keys


msgilligan:
The documentation for `mn create-acme-account` says "This command will either create a new account keypair for you or you can pass the account keypair that you have generated using the mn create-key or via openssl or other means in as a parameter."  (And the command does seem to behave according to the documentation.)

However the command help implies a key is always created.
 
### Steps to Reproduce

1. mn create-acme-account --help

### Expected Behaviour

* Help output should indicate an existing key can be used

### Actual Behaviour

```
Usage: mn create-acme-account [-fhvVx] -e=<email> [-k=<keyDir>] -n=<keyName> [-s=<keySize>]
                              (-u=<serverUrl> | --lets-encrypt-prod | --lets-encrypt-staging)
Creates a new account on the given ACME server
  -e, --email=<email>        Email address to create account with.
  -f, --force                Whether to overwrite existing files
  -h, --help                 Show this help message and exit.
  -k, --key-dir=<keyDir>     Custom location on disk to put the key to be used with this
                               account.
                               Default: src/main/resources
  -n, --key-name=<keyName>   Name of the key to be created
  -s, --key-size=<keySize>   Size of the key to be generated
                               Default: 4096
  -v, --verbose              Create verbose output.
  -V, --version              Print version information and exit.
  -x, --stacktrace           Show full stack trace when exceptions occur.
ACME server URL
      --lets-encrypt-prod    Use the Let's Encrypt prod URL.
      --lets-encrypt-staging Use the Let's Encrypt staging URL
  -u, --url=<serverUrl>      URL of ACME server to use
```

### Environment Information

- **Operating System**: macOS Big Sur
- **Micronaut Version:** 2.5.6
- **JDK Version:** 16

msgilligan:
I'm working on PR #850 to fix this.

ojebari-ma:
Can you please confirm that this is still an issue in the latest version of Micronaut?

PR #850, opened 2021-07-13
Fix mn create-acme-account usage

msgilligan:
“Name of the key to be used or created”

See Issue #849

This is a draft -- needs to be tested.


msgilligan:
This doesn't work and the following error is generated when the command is run:

```
Exception in thread "main" picocli.CommandLine$DuplicateOptionAnnotationsException: Option name '-n' is used by both field String io.micronaut.starter.cli.feature.acme.CreateAccount.keyName and field String io.micronaut.starter.cli.feature.acme.CreateKeyPair.keyName
```

msgilligan:
I've never used **picocli** before, this fix is probably easier for someone who knows it better.

PR #875, opened 2021-08-13
feat: rxjava2 feature

sdelamo:
See https://github.com/micronaut-projects/micronaut-core/pull/5976
See https://github.com/micronaut-projects/micronaut-core/pull/5977

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/other/HttpClient.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/other/HttpClient.java
@@ -21,7 +21,14 @@ import io.micronaut.starter.build.dependencies.Dependency;
 import io.micronaut.starter.feature.Category;
 import io.micronaut.starter.feature.DefaultFeature;
 import io.micronaut.starter.feature.Feature;
+import io.micronaut.starter.feature.FeatureContext;
 import io.micronaut.starter.feature.function.FunctionFeature;
+import io.micronaut.starter.feature.reactor.Reactor;
+import io.micronaut.starter.feature.reactor.ReactorHttpClient;
+import io.micronaut.starter.feature.rxjava.RxJavaThree;
+import io.micronaut.starter.feature.rxjava.RxJavaThreeHttpClient;
+import io.micronaut.starter.feature.rxjava.RxJavaTwo;
+import io.micronaut.starter.feature.rxjava.RxJavaTwoHttpClient;
 import io.micronaut.starter.options.Options;
 import jakarta.inject.Singleton;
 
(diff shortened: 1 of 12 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/other/HttpClient.java
```diff
@@ -30,12 +37,37 @@
 @Singleton
 public class HttpClient implements DefaultFeature {
 
+    private final RxJavaTwoHttpClient rxJavaTwoHttpClient;
+    private final RxJavaThreeHttpClient rxJavaThreeHttpClient;
```
ilopmar: Can we rename the classes to `RxJava2HttpClient`, RxJava3HttpClient` (and also the other one `RxJava1HttpClient`

starter-core/src/main/java/io/micronaut/starter/feature/other/HttpClient.java
```diff
@@ -30,12 +37,37 @@
...
+        if (featureContext.isPresent(RxJavaTwo.class)) {
+            featureContext.addFeature(rxJavaTwoHttpClient);
+        }
+        if (featureContext.isPresent(RxJavaThree.class)) {
+            featureContext.addFeature(rxJavaThreeHttpClient);
+        }
+        if (featureContext.isPresent(Reactor.class)) {
+            featureContext.addFeature(reactorHttpClient);
+        }
+    }
```
ilopmar: I think we should only allow to add one of these features. It doesn't make sense to have rxjava1, 2, 3 and reactor on the same app. Specially for RxJava.

ilopmar (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/other/HttpClient.java
sdelamo: It does not matter right?

starter-core/src/main/java/io/micronaut/starter/feature/other/HttpClient.java
sdelamo: But it should be part of a different PR right? If you add RxJava2 and you have other RxJava3 throw an exception.

starter-core/src/main/java/io/micronaut/starter/feature/other/HttpClient.java
ilopmar: No, we can name the classes whatever we want but I think we should be consistent with the names. We've always named them with numbers not with letters.

starter-core/src/main/java/io/micronaut/starter/feature/other/HttpClient.java
ilopmar: I don't see why. As part of adding a new feature we need to check the other features that are not compatible with the new one.

starter-core/src/main/java/io/micronaut/starter/feature/other/HttpClient.java
sdelamo: The class `RxJavaThree` already existed. That it is the reason I closed `RxJavaThreeHttpClient`. I can rename every class if you feel strong about this.

ilopmar (approved):

jameskleeh:
I changed the PR so the features for rxjava2-http-client get added from rxjava2 instead of from http-client. This is better because the rxjava2 related features and their logic are self contained

PR #876, opened 2021-08-13
feat: add reactive clients

sdelamo:
See https://github.com/micronaut-projects/micronaut-core/pull/5976
See https://github.com/micronaut-projects/micronaut-core/pull/5977


ilopmar (changes requested):
I miss the tests that only add the `xxxx-http-client` features and check that only that dependency is added but not the `micronaut-xxxx`. I mean, adding just `rxjava3-http-client` feature and get the artifact `micronaut-rxjava3-http-client`.

sdelamo:
`RxJavaOne to RxJava1 RxJavaThree to RxJava3`

> I miss the tests that only add the `xxxx-http-client` features and check that only that dependency is added but not the `micronaut-xxxx`. I mean, adding just `rxjava3-http-client` feature and get the artifact `micronaut-rxjava3-http-client`.

You cannot add `rxjava3-http-client` because they are are not visible.

ilopmar (approved):

jameskleeh:
I cherry picked this from 3.0.x

PR #877, opened 2021-08-16
feat: throw exception if multiple reactive libraries

sdelamo:


starter-core/src/main/java/io/micronaut/starter/feature/reactive/ReactiveFeatureValidator.java
```diff
@@ -0,0 +1,49 @@
...
+            throw new IllegalArgumentException("You cannot select more than one reactive library");
+        }
+    }
+
+    private static boolean moreThanOneReactiveFeature(Set<Feature> features) {
+        return features.stream()
+                .filter(f -> f instanceof ReactiveFeature)
+                .count() > 1;
+    }
+
```
ilopmar: Instead of creating a custom validator for this I think it's enough making `ReactiveFeature` implements/extends from `OneOfFeature`

ilopmar (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/reactive/ReactiveFeatureValidator.java
sdelamo: done via https://github.com/micronaut-projects/micronaut-starter/pull/877/commits/8bad0ac8c8deb53d0aac0d77720aea2e062eee4f

ilopmar (approved):

PR #902, opened 2021-09-13
Bump Gradle plugin to 2.0.5

melix:


ilopmar (approved):

melix:
Not quite sure why the build is still failing. In the beginning it was the Gradle plugin portal down but now the error seems unrelated.

melix:
I see that this also happens on other PRs so clearly not related to this change.

PR #918, opened 2021-09-28
Publish build scans on ge.micraunaut.io

melix:
The previous version was publishing build scans on the public scan
server. Now that we have a dedicated Gradle Enterprise instance it's
better to publish there since we will have more features available.


sdelamo (approved):

melix:
I think the failures are unrelated to this change.

ilopmar (approved):

PR #922, opened 2021-10-01
R2dbc improvements

### LINKED ISSUE #921: Add support for Oracle-R2DBC
### Feature description

Selecting the following features `data-r2dbc`, `jdbc-ucp`, `oracle` at https://micronaut.io/launch results in an error when clicking on the "PREVIEW" button -> "R2DBC is not yet supported by Oracle".

However there is an R2DBC implementation for Oracle -> https://github.com/oracle/oracle-r2dbc

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-starter/issues/921


ilopmar (changes requested):
Please add tests for the different drivers when no Liquibase/Flyway feature is selected. Also a new one (or modify it if exists) for Oracle R2DBC

graemerocher:
@ilopmar is that really necessary, the database tests are slow and duplicating them again for all databases without the migration feature seems overkill

ilopmar:
These should be unit test like all the existing ones for other features so they are fast. No need for new integration tests for this.

dstepanov:
I will add it

dstepanov:
Tests added

ilopmar (approved):

PR #923, opened 2021-10-02
gradlew.bat/mvnw.bat should NOT be executable

seregamorph:
If the gradle project is generated, the root directory listing has these gradle-wrapper related files:
```
ls -la demo/gradlew*
-rwxr-xr-x@ 1 morph  staff  8070 Oct  2 21:47 demo/gradlew
-rwxr-xr-x@ 1 morph  staff  2763 Oct  2 21:47 demo/gradlew.bat
```
The problem is that the second one, the bat file should not be executable, because it is auto-suggested in the shell to be executed, it raises inconvenience and requires manual fixing (via chmod). You can ensure and just list e.g. root directory of micronaut-starter:
```
ls -la micronaut-starter/gradlew*
-rwxr-xr-x  1 morph  staff  8070 Oct  2 22:45 micronaut-starter/gradlew
-rw-r--r--  1 morph  staff  2763 Oct  2 22:45 micronaut-starter/gradlew.bat
```
The first file for *nix has the executable flag, the second file for Windows - does not.

This PR addresses the issue.

How to check: e.g. run Application in `starter-web-netty` module, then execute
`curl "http://localhost:8080/create/DEFAULT/example?build=gradle" -o example.zip` and unpack the archive.

The same for maven wrapper bat executable.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/Gradle.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/Gradle.java
@@ -60,7 +60,7 @@ public class Gradle implements BuildFeature {
         generatorContext.addTemplate("gradleWrapperJar", new BinaryTemplate(WRAPPER_JAR, classLoader.getResource(WRAPPER_JAR)));
         generatorContext.addTemplate("gradleWrapperProperties", new URLTemplate(WRAPPER_PROPS, classLoader.getResource(WRAPPER_PROPS)));
         generatorContext.addTemplate("gradleWrapper", new URLTemplate("gradlew", classLoader.getResource("gradle/gradlew"), true));
-        generatorContext.addTemplate("gradleWrapperBat", new URLTemplate("gradlew.bat", classLoader.getResource("gradle/gradlew.bat"), true));
+        generatorContext.addTemplate("gradleWrapperBat", new URLTemplate("gradlew.bat", classLoader.getResource("gradle/gradlew.bat"), false));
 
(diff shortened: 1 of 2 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/Gradle.java
```diff
@@ -60,7 +60,7 @@ public void apply(GeneratorContext generatorContext) {
         generatorContext.addTemplate("gradleWrapperJar", new BinaryTemplate(WRAPPER_JAR, classLoader.getResource(WRAPPER_JAR)));
         generatorContext.addTemplate("gradleWrapperProperties", new URLTemplate(WRAPPER_PROPS, classLoader.getResource(WRAPPER_PROPS)));
         generatorContext.addTemplate("gradleWrapper", new URLTemplate("gradlew", classLoader.getResource("gradle/gradlew"), true));
-        generatorContext.addTemplate("gradleWrapperBat", new URLTemplate("gradlew.bat", classLoader.getResource("gradle/gradlew.bat"), true));
+        generatorContext.addTemplate("gradleWrapperBat", new URLTemplate("gradlew.bat", classLoader.getResource("gradle/gradlew.bat"), false));
```
seregamorph, line 63: last argument that is changed to `false` - `executable`

jameskleeh:
Does this not impact Windows users? They would now have to manually set the file to be executable, correct?

seregamorph:
@jameskleeh no, it does not affect Windows users. Windows does not have executable flag in file system. `java.io.File.setExecutable()` is no-op for Windows machines.
>They would now have to manually set the file to be executable

No, they will not notice anything

jameskleeh:
Thanks

PR #927, opened 2021-10-06
Use more appropriate default memory setting for lambda

graemerocher:


starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
```diff
@@ -159,6 +159,15 @@ azurefunctions {
 }
 }
 
+@if (features.contains("aws-lambda")) {
```
sdelamo, line 162: this should only be present if `graalvm` feature is present as well ?

graemerocher: The graalvm feature should always be there. For Gradle you don’t need it to build native images

ilopmar: What about the equivalent configuration for Maven users?

graemerocher: yeah I agree that needs to be place, but looking through the current `pom.xml.raw` there are many things missing from Maven with regards to the serverless support since nobody is paying attention to it much

I was going to add it but then saw pretty much everything related to serverless is not implemented for Maven as far as I can see

graemerocher: Note that today the experience is for new users that if they just choose the defaults and create an AWS lambda app it by default will not be able to run (fails on the 3rd or 4th request) unless these memory settings are in place.

If we are happy for that to be the developer experience for Lambda users then we can close this PR

sdelamo: I am creating a PR against this PR, please give me some time. 

We are all on the same team. Nobody, is happy about a bad experience for users. We will merge this PR.

sdelamo:
@graemerocher I have moved the logic to https://github.com/micronaut-projects/micronaut-starter/pull/928 I have asked for a review in that PR. 

@ilopmar @graemerocher I have created a separate issue for Maven https://github.com/micronaut-projects/micronaut-starter/issues/929

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
graemerocher: @sdelamo just pinging that the other PR you created has test failures, are we going to leave this as is for longer?

sdelamo: Let me check the other PR.

PR #932, opened 2021-10-07
Use JDK17 on CI

ilopmar:
This should fix all the tests failing like this one:

```
io.micronaut.starter.cli.feature.database.CreateRepositorySpec test creating a repository - groovy FAILED

  java.lang.IllegalArgumentException: Unsupported JDK version: 16. Supported values are [8, 11, 17]
```

Because JDK16 is not supported anymore in Starter.


ilopmar:
This fails because of the fix in #931. I've added this commit to that other PR.

PR #941, opened 2021-10-13
Update camunda and camunda-external-worker to versions compatible wit…

tobiasschaefer:
…h Micronaut 3.1.0

I'm the lead developer of the Micronaut Camunda Integration Project at https://github.com/camunda-community-hub/micronaut-camunda-bpm. Please merge this PR.

### DIFF
--- a/starter-core/src/main/resources/pom.xml
+++ b/starter-core/src/main/resources/pom.xml
@@ -15,17 +15,17 @@
         <dependency>
             <groupId>org.camunda.bpm.assert</groupId>
             <artifactId>camunda-bpm-assert</artifactId>
-            <version>10.0.0</version>
+            <version>13.0.0</version>
         </dependency>
         <dependency>
             <groupId>info.novatec</groupId>
             <artifactId>micronaut-camunda-bpm-feature</artifactId>
-            <version>2.0.0</version>
+            <version>2.1.0</version>
         </dependency>
         <dependency>
             <groupId>info.novatec</groupId>
             <artifactId>micronaut-camunda-external-client-feature</artifactId>
-            <version>2.0.0</version>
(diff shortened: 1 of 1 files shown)


sdelamo (approved):

ilopmar:
@tobiasschaefer just to be clear the previous versions from this PR doesn't work with Micronaut 3.1.x?

sdelamo:
Thanks @tobiasschaefer !

sdelamo:
> @tobiasschaefer just to be clear the previous versions from this PR doesn't work with Micronaut 3.1.x?


I think they work but this new versions are built with Micronaut 3.1.x

https://twitter.com/toschaef/status/1448146356607111169

ilopmar:
I'm asking because merging this PR in `3.1.x` branch means upgrading to a major version in a patch release. Maybe we should target 3.2.x instead.

sdelamo:
I don't think we have to be "semantic version" strict about feature dependency versions in starter. 

Their new version is ready for JDK 17 and Micronaut 3.1. I think the best experience for users creating 3.1. and selecting camunda is to get this version of the library.

ilopmar:
Ok, feel free to merge it :-)

tobiasschaefer:
Yes, Micronaut 3.1.x will also work with our 2.0.x release but it would be much better to provide the current 2.1.x release which has been built with Micronaut 3.1.

I'd appreciate it, if you wouldn't look to closely on the semantic version numbers. As an open source project with limited resources and not fully integrated into the Micronaut Release process we would always be on step behind if you don't merge it. That would from my point of view not be the best experience for the user.

Please merge :-)

ilopmar (approved):

PR #945, opened 2021-10-18
Bump version of Coherence.

ghost:
Integrate latest version of Coherence CE.


starter-core/src/main/java/io/micronaut/starter/feature/coherence/CoherenceFeature.java
```diff
@@ -32,7 +32,7 @@
 @Singleton
 public class CoherenceFeature implements Feature {
 
-    public static final String COHERENCE_VERSION = "21.06.1";
+    public static final String COHERENCE_VERSION = "21.06.2";
```
ilopmar, line 35: If you put the coordinates and version in `pom.xml` (https://github.com/micronaut-projects/micronaut-starter/blob/3.1.x/starter-core/src/main/resources/pom.xml) we'll get automatic PRs to update the version from dependabot.

PR #946, opened 2021-10-19
Correct oracle r2dbc driver group

dstepanov:


graemerocher:
Is there a test for this?

dstepanov:
Only for checking if the driver's name is present.

PR #953, opened 2021-10-27
Update camunda feature to latest release

tobiasschaefer:
I'm the lead developer of the Micronaut Camunda Integration Project at https://github.com/camunda-community-hub/micronaut-camunda-bpm. Please merge this PR.

### DIFF
--- a/starter-core/src/main/resources/pom.xml
+++ b/starter-core/src/main/resources/pom.xml
@@ -20,7 +20,7 @@
         <dependency>
             <groupId>info.novatec</groupId>
             <artifactId>micronaut-camunda-bpm-feature</artifactId>
-            <version>2.1.0</version>
+            <version>2.2.0</version>
         </dependency>
         <dependency>
             <groupId>info.novatec</groupId>


ilopmar:
This should target branch `3.2.x`

tobiasschaefer:
Hi @ilopmar ,

thanks for looking into this PR.

Should this target 3.2.x because our dependency update is a minor update, not only patch?

If so: I'd appreciate it, if you wouldn't look too closely on the semantic version numbers. As an open source project with limited resources and not fully integrated into the Micronaut Release process we would always be on step behind if you don't merge it. That would from my point of view not be the best experience for the user. This pull request will definitely not break anything.

A similar pull request was merged in #941 and I'd really appreciate if you could merge this one too.

When will 3.2x be released?

ilopmar:
This is what we follow regarding versioning in modules and include it in core's BOM: https://github.com/micronaut-projects/micronaut-core/wiki/Module-Upgrade-Strategy

Giving it a second look I've realized that the micronaut-camunda module is not in our BOM. So this new version upgrade would only affect to new users creating applications with Launch.
I'm going to merge the PR.

3.2.0 will be release before the end on November.

ilopmar (approved):

tobiasschaefer:
Thanks @ilopmar 

Yes, this only affects newly created applications - we're not listed in the BOM.

PR #954, opened 2021-11-03
feat: add kubernetes clients features

pgressa:


starter-core/src/main/java/io/micronaut/starter/feature/k8s/KubernetesReactorClient.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public String getThirdPartyDocumentation() {
+        return "https://github.com/kubernetes-client/java/wiki";
+    }
+
+    @Override
+    public void processSelectedFeatures(FeatureContext featureContext) {
+        if (featureContext.isPresent(KubernetesClient.class)) {
+            featureContext.exclude(feature -> feature instanceof KubernetesClient);
```
jameskleeh, line 76: Are the two clients incompatible with eachother? If so this seems like a use case for `OneOfFeature` instead of silently removing the non reactive client

starter-core/src/main/java/io/micronaut/starter/feature/k8s/KubernetesReactorClient.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.addDependency(Dependency.builder()
```
jameskleeh: This should probably check if micronaut reactor or micronaut reactor http client is part of the features since this would be redundant in that case

starter-core/src/main/java/io/micronaut/starter/feature/k8s/KubernetesRxJava2Client.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public void processSelectedFeatures(FeatureContext featureContext) {
+        if (featureContext.isPresent(KubernetesClient.class)) {
+            featureContext.exclude(feature -> feature instanceof KubernetesClient);
+        }
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.addDependency(Dependency.builder()
```
jameskleeh: Same comment as above but for micronaut rxjava2

jameskleeh (changes requested):
Is there no kubernetes-rxjava3-client ?

starter-core/src/main/java/io/micronaut/starter/feature/k8s/KubernetesReactorClient.java
pgressa: They are compatible.

pgressa:
@jameskleeh no rx3 yet, but good idea, i'll create an issue

starter-core/src/main/java/io/micronaut/starter/feature/k8s/KubernetesReactorClient.java
pgressa: true, will add

starter-core/src/main/java/io/micronaut/starter/feature/k8s/KubernetesReactorClient.java
pgressa: It's just it's not needed to reference kubernetes client, since the dependency is part of the reactive clients

graemerocher:
@jameskleeh I don't think the lack of an rxjava3 client should hold this up

graemerocher (approved):

jameskleeh:
Good to merge once the build passes

pgressa:
@jameskleeh build passed, except status

PR #955, opened 2021-11-03
feat: add oracle cloud micrometer

### LINKED ISSUE #737: Add oracle-cloud micrometer and atp modules

pgressa:
Fixes #737

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/micrometer/AppOptics.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/micrometer/AppOptics.java
@@ -29,23 +29,19 @@ public class AppOptics extends MicrometerFeature {
     }
 
     @Override
-    public String getName() {
-        return "micrometer-appoptics";
-    }
-
-    @Override
     public String getDescription() {
         return "Adds support for Micrometer metrics (w/ AppOptics reporter)";
     }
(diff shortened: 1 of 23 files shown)


graemerocher (approved):

jameskleeh:
Refactoring to use a method for the dependency artifact and group doesn't make sense to me because the parent class isn't using the method

Perhaps having `getDependencyGroupName` there for reuse makes sense but the conflation of the feature name with the artifact name doesn't seem right to me

Edit: The dependency group name can just be a constant in the parent class as well. No need for a method

pgressa:
@jameskleeh it is there just for the sake of tests. Yes it's very ugly, but if I want to use MicrometerFeature then either I would have to tweak the tests someway to say like if it's oracle micrometer, then use different dependency [name&&group], since the oracle micrometer has completely different format of the dependency. 

What I can do is to automatically add the micronaut micrometer dependency in the parent class, so the methods will be used there. 

OR

make the tests uglier

OR

not use MicrometerFeature at all.. 

I don't mind to do any, choose what suites you more.

graemerocher:
@jameskleeh any feedback on a path forward?

jameskleeh:
@pgressa I think the dependency should be added in the parent class

PR #958, opened 2021-11-05
Upgrade the Gradle plugin to 3.0.0-M1

melix:
Let's see what tests fail!

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
@@ -109,9 +109,15 @@ java {
 }
 }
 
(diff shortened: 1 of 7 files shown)


ilopmar (changes requested):
Please undo all the indents you made. I agree that they make the rocker template more readable but those indents will be in the final `build.gradle` created.

melix:
No, they won't. Now there will be a consistent `tasks { ... }` block independently of using Groovy or Kotlin DSL. Before, if it was Groovy, then there was no indent but tasks were configured at the top level. It was absent if there was no task customization. If it was Kotlin DSL, there was a top level `tasks` block and the tasks were configured within, but it was _present_ (and empty) if there was no task customization. Now in both cases we use the `tasks { ... }` block and it's only present if there are tasks customizations.

ilopmar:
Oh, you are right, sorry for the noise

ilopmar (approved):

PR #962, opened 2021-11-08
Disable Gradle Toolchains support

ilopmar:


melix:
I am not sure it's a good idea. Folks may need toolchain support for other reasons than Micronaut itself. By disabling toolchain support we just break all cases where it works fine.

jameskleeh:
It wasn't enabled before so this is simply not adding the support rather than breaking anything. Given we can't specify a specific version adding this has the potential for more harm than good, in my opinion. If users want it they can enable it after creating the app, as it was before.

melix:
Strictly speaking this isn't true: toolchain support was there for anyone doing:

```
java {
   toolchain {
      languageVersion.set(...)
}
```

however I do agree that because it's a template, the risk is lower. At the same time, because it's a template (reverse argument), it means that the risk of being surprised by the behavior is lower.

ilopmar:
I don't agree with your last sentence :stuck_out_tongue:. Every user that has used our plugin before and is not using toolchains will be surprised by this new behavior (as I did).

jameskleeh:
The docs specifically say "If your application depends on a specific GraalVM version, you will have to disable automatic detection like explained below.". Micronaut 3.2 applications do require a specific version of GraalVM, so therefore it should be disabled. 

Ideally I would prefer it to be disabled in the plugin itself, however given users have to do manual work to upgrade they can make that decision for themselves. For new apps we should do what we recommend and in this case that means disabling the support.

jameskleeh (approved):

melix:
The toolchain support is _not_ specific to a plugin, so it cannot be disabled by a plugin: it's a generic feature of Gradle which works well and solves real problems.  As a long time Gradle users I would personally be surprised that Micronaut disables it, since it would prevent me typically from building apps which depend on multiple JDKs, test on different JVMs from a single build, etc., so the argument "folks have used Micronaut before" doesn't stand from my POV. There's in absolute no reason why you would put building native-images with GraalVM more important than making sure your application builds reproducibly on a specific JDK, which is what toolchains support provides.

I agree, however, that there *is* one issue with GraalVM in that you cannot select a specific version of GraalVM itself, which is why we document how to workaround this issue. For reference: https://github.com/gradle/gradle/issues/18290

I have a meeting tomorrow to determine if the Gradle team considers the issue above will be fixed.

jameskleeh:
@melix If and when that is fixed we can revisit this

PR #972, opened 2021-11-19
TOML configuration feature

yawkat:
Draft of the TOML config feature.

By default, config properties are written property-like using normal dotted keys:

```toml
foo.bar = 1
foo.baz = 2
```

Using the new `config.registerTable("foo")` method, this can be changed to the semantically identical, but prettier, table-based variant:

```toml
[foo]
bar = 1
baz = 2
```

The `registerTable` call has no impact for other config formats, it's purely supplemental metadata for toml.


yawkat:
Instead of requiring each feature to register their tables manually, I've added an algorithm that tries to suggest tables where possible. I've tested this with all the features at the same time. The resulting config files aren't perfect, but they look reasonable enough: https://gist.github.com/yawkat/73e904ab83de0c7a275667517590eeb1

starter-core/src/main/java/io/micronaut/starter/feature/config/Toml.java
```diff
@@ -0,0 +1,71 @@
...
+import io.micronaut.starter.build.dependencies.Dependency;
+import io.micronaut.starter.feature.FeaturePhase;
+import io.micronaut.starter.template.Template;
+import io.micronaut.starter.template.TomlTemplate;
+import jakarta.inject.Singleton;
+
+import java.util.function.Function;
+
+@Singleton
+public class Toml implements ConfigurationFeature {
```
sdelamo, line 31: I think you have to override `getCategory()`

starter-core/src/test/groovy/io/micronaut/starter/template/TomlTemplateSpec.groovy
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    def 'simple'() {
+        given:
+        def config = new Configuration('', '', '')
+        config.put('foo.bar', 4)
+        config.put('foo.baz', ['foo': 123, 'bar': [1, 2, ['name': 'fox']]])
+
+        expect:
+        toml(config) == '''foo.bar = 4
```
sdelamo: I would recommend to replace:

```groovy
        toml(config) == '''foo.bar = 4
```
with: 

```
        toml(config) == '''\
foo.bar = 4
```

see: https://groovy-lang.org/syntax.html#_triple_single_quoted_string

sdelamo (changes requested):
I think we should at least have a functional test where we generate an app with toml and see that test execute fine.

graemerocher:
Yeah a bit more testing would be good, also with a more complex example

starter-core/src/main/java/io/micronaut/starter/feature/config/Toml.java
yawkat: looks like `ConfigurationFeature` already overrides it, with `Category.CONFIGURATION`

yawkat:
@sdelamo how do I do that? Is there an example for such a full test?

sdelamo:
> @sdelamo how do I do that? Is there an example for such a full test?

check tests in `test-core` submodule.

graemerocher (approved):

sdelamo (approved):

PR #982, opened 2021-11-24
Merge up 3.1.x into 3.2.x

sdelamo:
@jameskleeh I think you forgot to do a merge up from 3.1.x to 3.2.x prior to the release.


ilopmar:
This PR shouldn't be merged because if we do it we'll loss track of all individual commits. This PR only should be used to check that the build passes.

sdelamo:
@jameskleeh  why is this PR closed?  @ilopmar are you doing the merge up instead?

ilopmar:
Done. Merge up and pushed the changes.

PR #1013, opened 2021-12-23
Fix Gradle dependency snippet for POMs when Gradle Kotlin DSL selected

scprek:


scprek:
@graemerocher looks like the checks have not re-run after a quick fix I added. Do you need to approve again?

PR #1014, opened 2021-12-24
Update knative template so that it will work with latest Knative version

Deviad:
Screenshot 2021-12-25 at 16 47 50
I have removed a few properties from the YAML config because they seem not to be supported anymore in the new version of Knative.


Deviad:
Hello, any news on this one?

PR #1019, opened 2022-01-06
Use the Gradle build cache to make builds faster

melix:
This is an attempt to make the Micronaut Starter builds faster when executing tests. The idea is that in most cases, the _build under test_ is always the same for a particular test: when a change is made to Micronaut Starter, unless it's a version change for a dependency like Micronaut core or Gradle, most of the tests should have nothing changed.

Therefore, we're adding the remote cache to the _builds under test_: if they use the same classpath and same build scripts, then they would be hopefully found in the cache and the test would execute faster.

This is an experiment, to see if this really works in practice, since the optimization depends on the build cache size and the repetition of builds.


sdelamo:
I have changed this to use our internal gradle plugin.

timyates (approved):

PR #1021, opened 2022-01-09
support for views-jte

edward3h:
I added [JTE support in micronaut-views](https://github.com/micronaut-projects/micronaut-views/pull/320).
This adds it to Launch.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/view/JTE.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/view/JTE.java
@@ -29,6 +29,7 @@ import java.nio.charset.StandardCharsets;
 
 @Singleton
 public class JTE implements ViewFeature, MicronautServerDependent {
+    private final static String JTE_SRC_DIR = "src/main/jte";
 
     @Override
     public String getName() {
@@ -59,12 +60,11 @@ public class JTE implements ViewFeature, MicronautServerDependent {
     public void apply(GeneratorContext generatorContext) {
         generatorContext.addDependency(Dependency.builder()
                 .groupId("io.micronaut.views")
-                .lookupArtifactId("micronaut-views-jte")
+                .artifactId("micronaut-views-jte")
                 .compile());
-        String jteSrcDir = jteSrcDir(generatorContext);
         generatorContext.addBuildPlugin(GradlePlugin.builder()
                 .id("gg.jte.gradle")
-                .extension(new RockerWritable(gradlePluginJTE.template(jteSrcDir)))
+                .extension(new RockerWritable(gradlePluginJTE.template(JTE_SRC_DIR)))
                 .lookupArtifactId("jte-gradle-plugin")
                 .build());
         String mavenPluginArtifactId = "jte-maven-plugin";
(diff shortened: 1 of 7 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/view/JTE.java
```diff
@@ -0,0 +1,90 @@
...
+                .extension(new RockerWritable(mvnPluginJTE.template(coordinate.getGroupId(),
+                        coordinate.getArtifactId(),
+                        coordinate.getVersion(),
+                        jteSrcDir)))
+                .build());
+        generatorContext.addTemplate("exampleJte", new RockerTemplate(jteSrcDir + "/example.jte", exampleJTE.template()));
+    }
+
+    private String jteSrcDir(GeneratorContext generatorContext) {
+        String path = generatorContext.getConfiguration().getPath();
```
jameskleeh: You shouldn't be doing path calculation, just specify the full path.

`src/main/resources/jte/example.jte` for example

starter-core/src/main/resources/pom.xml
```diff
@@ -157,5 +157,21 @@
             <artifactId>plugin</artifactId>
             <version>5.12.0</version>
         </dependency>
+        <dependency>
+            <!-- FIXME: this can be removed once the micronaut-bom references version 3.1.2 or higher -->
+            <groupId>io.micronaut.views</groupId>
+            <artifactId>micronaut-views-jte</artifactId>
+            <version>3.1.2</version>
```
jameskleeh: This can be removed

starter-core/src/main/java/io/micronaut/starter/feature/view/JTE.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public String getMicronautDocumentation() {
+        return "https://micronaut-projects.github.io/micronaut-views/latest/guide/#jte";
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.addDependency(Dependency.builder()
+                .groupId("io.micronaut.views")
+                .lookupArtifactId("micronaut-views-jte")
```
jameskleeh: This should just be artifactId

jameskleeh (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/view/JTE.java
edward3h: changed

edward3h:
I have some failing tests now - will fix.

starter-core/src/main/resources/pom.xml
edward3h: My build is currently pulling micronaut-bom-3.2.8-SNAPSHOT and this still refers to micronaut-views-bom:3.0.4, which doesn't include micronaut-views-jte.

Is there something else I need to do to pick up the newer version of micronaut-views?

sdelamo:
Closed via https://github.com/micronaut-projects/micronaut-starter/pull/1045

PR #1035, opened 2022-01-21
Add Netty max order config / fix config nesting

graemerocher:
This PR does two things:

1. Adds Netty max order config by default so that folks creating new apps and benchmarking get equivalent memory behaviour to rival frameworks
2. Adds the ability to define nested config values more easily since the Configuration type and concept seemed to not support YAML nesting correctly


graemerocher:
Test failures are unrelated

jameskleeh:
The failure seems related. I'm not seeing it in any other builds

graemerocher:
I merged up and it identified a new test failure which I fixed. But the servlet related failures are not related to this PR. They are occurring on @sdelamo's PR as well https://github.com/micronaut-projects/micronaut-starter/runs/4949356006?check_suite_focus=true#step:6:3027

PR #1041, opened 2022-01-25
Add Micronaut Data MongoDB preview feature

graemerocher:


graemerocher:
Note versions are hard coded until BOM is upgraded

dstepanov (approved):

starter-core/src/main/java/io/micronaut/starter/feature/database/DataMongo.java
```diff
@@ -0,0 +1,100 @@
...
+                .groupId("io.micronaut.data")
+                .artifactId("micronaut-data-document-processor")
+                .version("3.3.0-M1")
+        );
+        generatorContext.addDependency(
+            Dependency.builder()
+                .compile()
+                .groupId("io.micronaut.data")
+                .artifactId("micronaut-data-mongodb")
+                .version("3.3.0-M1")
```
ilopmar, line 77: The version should be on https://github.com/micronaut-projects/micronaut-starter/blob/3.3.x/starter-core/src/main/resources/pom.xml

graemerocher: The plan is for this version to move to the BOM once 3.3.0 is final so I am not sure of the need?

ilopmar: Just in case we forget to remove the hardcoded version here. At least we could get a dependabot PR, remember that the version is in the BOM and then remove this. If we don't remove the version here we will always generate apps with it, even if micronaut-data-mongodb is added to the bom

graemerocher: when the feature is taken out of preview this feature will need to be modified to remove the `isPreview()` method, at that point it will be fairly obvious that the -M1 hard coded versions should be removed.

On the other hand it is more likely to be forgotten if these versions are instead shifted to another non-visible file that won't be updated until a dependabot PR arrives that we would have to pay attention to.

ilopmar: :+1:

PR #1047, opened 2022-01-27
gradle enterprise

sdelamo:
This add support for adding the Gradle Enterprise plugin. Useful, if you want to use starter-core in your company and generate all your apps pushing to a particular server.


sdelamo:
@alextu It would be great to support this for maven as well. Not sure if you can help with it.

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/settingsGradle.rocker.raw
```diff
@@ -1,8 +1,21 @@
...
 @import io.micronaut.starter.build.gradle.GradleBuild
-
+@import io.micronaut.starter.build.gradle.GradlePlugin
+@import io.micronaut.starter.build.gradle.GradleDsl
 @args (Project project, GradleBuild gradleBuild)
 
+@if(!gradleBuild.getSettingsPlugins().isEmpty()) {
+plugins {
+@for (GradlePlugin gradlePlugin : gradleBuild.getSettingsPlugins()) {
+if (gradleBuild.getDsl() == GradleDsl.KOTLIN) {
```
timyates: should be `@if (gradleBuild.getDsl() == GradleDsl.KOTLIN)` (missing `@`)

timyates (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/settingsGradle.rocker.raw
```diff
@@ -1,8 +1,21 @@
...
+    id("@gradlePlugin.getId()") @(gradlePlugin.getVersion() != null ? "version(\"" + gradlePlugin.getVersion() + "\")" : "")
+} else {
+    id "@gradlePlugin.getId()" @(gradlePlugin.getVersion() != null ? "version \"" + gradlePlugin.getVersion() + "\"" : "")
+}
+}
+}
+}
+
```
timyates: NIT: Another CRLF here makes the file look nicer

timyates (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/settingsGradle.rocker.raw
timyates: fix pushed

timyates (approved):

alextu:
> @alextu It would be great to support this for maven as well. Not sure if you can help with it.

Thanks for tagging me, I'll check if it's something we can take

starter-core/src/test/groovy/io/micronaut/starter/feature/build/gradle/MicronautGradleEnterpriseSpec.groovy
```diff
@@ -0,0 +1,49 @@
...
+
+        then:
+        settings.contains('id("com.gradle.enterprise") version("3.8.1")')
+        !settings.contains("allowUntrustedServer = true")
+        settings.contains('gradleEnterprise {')
+        settings.contains('server = "https://ge.micronaut.io"')
+        settings.contains('buildScan {')
+        settings.contains('termsOfServiceUrl = "https://gradle.com/terms-of-service"')
+        settings.contains('termsOfServiceAgree = "yes"')
+        if (buildTool == BuildTool.GRADLE_KOTLIN) {
```
etiennestuder: TOS agreement is only needed when publishing to scans.gradle.com. When configuring a GE server via `server = `, there is no need to accept the TOS since the customer running that instance has already agreed to the TOS separately.

etiennestuder (changes requested):
@sdelamo It is of course your call, but if you could address the feedback I left re: TOS in the settings.gradle, that would be great.

starter-core/src/test/groovy/io/micronaut/starter/feature/build/gradle/MicronautGradleEnterpriseSpec.groovy
sdelamo: I have done the change.

PR #1048, opened 2022-01-27
Build cache

sdelamo:
This relates to: https://github.com/micronaut-projects/micronaut-starter/pull/1019


starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildCache.rocker.raw
```diff
@@ -0,0 +1,34 @@
...
+@if(dsl == GradleDsl.KOTLIN) {
+buildCache {
+    remote<HttpBuildCache> {
+        url = uri("@(config.getUrl())")
+@if(config.push()) {
+        isPush = true
+}
+@if(config.getCredentials() != null) {
+        credentials {
+            username = "@(config.getCredentials().getUsername())"
```
melix: Probably we want to recommend that the username "value" is actually something like `System.getProperty("...")`, so that the script doesn't contain the full username/password. In this case it probably also makes sense to escape the value so that quotes are properly rendered.

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/MicronautBuildCache.java
```diff
@@ -0,0 +1,96 @@
...
+            generatorContext.addBuildPlugin(GradlePlugin.builder()
+                    .settingsExtension(new RockerWritable(buildCache.template(generatorContext.getBuildTool().getGradleDsl().orElse(GradleDsl.GROOVY), this)))
+                    .build());
+        }
+    }
+
+    @Override
+    @Nullable
+    public Credentials getCredentials() {
+        String cacheUsername = System.getProperty("ge.cache.username");
```
melix: I am actually wondering if we should do this. It would probably make more sense that the _build script_ reads the property.

melix (changes requested):
I think we should change how credentials are injected. The tests can determine whether to inject the feature or not based on the presence of the system property, but they could use a _different_ one in the test itself.

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildCache.rocker.raw
sdelamo: See https://github.com/micronaut-projects/micronaut-starter/pull/1048/commits/8e03dcb8ed46edbd0bbd2b9ee85f911f294b47d6 I think that should resolve your concerns. If it does, please resolve this conversation.

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/MicronautBuildCache.java
sdelamo: See https://github.com/micronaut-projects/micronaut-starter/pull/1048/commits/8e03dcb8ed46edbd0bbd2b9ee85f911f294b47d6 I think that should resolve your concerns. If it does, please resolve this conversation.

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/MicronautBuildCache.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    @NonNull
+    public String getUrl() {
+        return "https://ge.micronaut.io/cache/";
+    }
+
+    @Override
+    public boolean push() {
+        return true;
```
timyates, line 53: Is this for CI?  I assume we only want to push from CI?  Is this a sensible default?  Or should we default to some property that only exists on CI?

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/GradleBuildCache.java
```diff
@@ -0,0 +1,72 @@
...
+import io.micronaut.starter.application.ApplicationType;
+import io.micronaut.starter.application.generator.GeneratorContext;
+import io.micronaut.starter.build.gradle.GradleDsl;
+import io.micronaut.starter.build.gradle.GradlePlugin;
+import io.micronaut.starter.feature.Category;
+import io.micronaut.starter.feature.Feature;
+import io.micronaut.starter.feature.build.gradle.templates.buildCache;
+import io.micronaut.starter.template.RockerWritable;
+
+public abstract class GradleBuildCache implements Feature, BuildRemoteCacheConfiguration {
```
melix, line 29: I am confused. You're not overriding `isVisible()`, which defaults to `true`. So how does this feature gets hidden?

sdelamo: I am overriding it in `MicronautBuildCache` which is the only `@Singleton` implementation. `GradleBuildCache` is an abstract class.

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/MicronautBuildCache.java
sdelamo: Yes, it would be used in `CI`. Please, note that `MicronautBuildCache` is not `visible` see `isVisible() { return false; }`. It would be used by our generated projects on Ci. 

I think we needed as `push = true` according to https://github.com/micronaut-projects/micronaut-starter/pull/1019/files See `maybeaddGradleBuldCacheGradleConfiguration`method. Maybe @melix can clarify this.

timyates: Cool!  I was just worried people might push from their local machines, but I guess without credentials it would fail 👍

timyates (approved):

Issue #897, opened 2021-08-26
Add feature(s) for swagger-ui


opticyclic:
https://micronaut.io/launch doesn't create a swagger-ui view when selecting the OpenAPI option

The docs say you can enable the view with a `openapi.properties` file containing `swagger-ui.enabled=true`
https://micronaut-projects.github.io/micronaut-openapi/latest/guide/index.html#viewsGenerationWithPropertiesFile
https://micronaut-projects.github.io/micronaut-openapi/latest/guide/index.html#propertiesFileConfiguration

However, adding this to the generated project doesn't create a view at http://localhost:8080/swagger-ui/

Ideally the generated project should contain this since I'd expect people adding OpenAPI to their starter project would like a view to check it is working.

Adding a comment in the generated `openapi.properties` file or the README explaining how to enable/disable it would work for the people that want to see it and also for the people who don't.

If it can't be added it to the generator it would be useful to link to a working example from the docs.

ilopmar:
Those are two different things, one is create the swagger-ui automatically (or redoc or rapidoc) and another thing is exposing it.

If you take a look at the docs you linked: https://micronaut-projects.github.io/micronaut-openapi/latest/guide/index.html#swaggerui

![image](https://user-images.githubusercontent.com/559192/132226582-8dfd85dd-0836-44b1-a0c2-d5542e1d05b5.png)

ilopmar:
So, for me this works as expected. Another different thing would be to add new features to Launch to configure automatically `swagger-ui`, `rapidoc` and `redoc` views to expose the generated views automatically (also taking security into account)

pgressa:
Nemanja Mikic works on this stuff here, but he can't assign the issue yet

altro3:
@sdelamo as I see, it's already done

PR #1053, opened 2022-01-28
Add swagger-ui feature

n0tl3ss:
Solution for #897 

The swagger-ui is exposed by default adding this feature.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/other/template/openApiProperties.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/other/template/openApiProperties.rocker.raw
@@ -1,0 +1,6 @@
+swagger-ui.enabled=true
+redoc.enabled=true
+rapidoc.enabled=true
+rapidoc.bg-color=#14191f
+rapidoc.text-color=#aec2e0
+rapidoc.sort-endpoints-by=method
--- a/starter-core/src/main/java/io/micronaut/starter/feature/other/template/openApiProperties.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/other/template/openApiProperties.rocker.raw
@@ -1,6 +1,6 @@
 swagger-ui.enabled=true
-redoc.enabled=true
-rapidoc.enabled=true
+redoc.enabled=false
+rapidoc.enabled=false
 rapidoc.bg-color=#14191f
 rapidoc.text-color=#aec2e0
 rapidoc.sort-endpoints-by=method
--- a/starter-core/src/test/groovy/io/micronaut/starter/feature/other/SwaggerUISpec.groovy
+++ b/starter-core/src/test/groovy/io/micronaut/starter/feature/other/SwaggerUISpec.groovy
@@ -49,7 +49,6 @@ class SwaggerUISpec extends ApplicationContextSpec  implements CommandOutputFixt
         ctx.configuration.get('micronaut.router.static-resources.swagger.mapping') == "/swagger/**"
         ctx.configuration.get('micronaut.router.static-resources.swagger-ui.paths') == "classpath:META-INF/swagger/views/swagger-ui"
         ctx.configuration.get('micronaut.router.static-resources.swagger-ui.mapping') == "/swagger-ui/**"
-        ctx.configuration.get('micronaut.router.static-resources.swagger-ui.mapping') == "/swagger-ui/**"
 
         List<Map<String, String>> swaggerSec = ctx.configuration.get('micronaut.security.intercept-url-map') as List<Map<String, String>>
 
--- a/starter-core/src/test/groovy/io/micronaut/starter/feature/other/SwaggerUISpec.groovy
+++ b/starter-core/src/test/groovy/io/micronaut/starter/feature/other/SwaggerUISpec.groovy
@@ -1,0 +1,74 @@
+package io.micronaut.starter.feature.other
+
+import io.micronaut.starter.ApplicationContextSpec
+import io.micronaut.starter.application.ApplicationType
+import io.micronaut.starter.application.generator.GeneratorContext
+import io.micronaut.starter.feature.Features
+import io.micronaut.starter.fixture.CommandOutputFixture
+
+class SwaggerUISpec extends ApplicationContextSpec  implements CommandOutputFixture {
+
+    void "test swagger-ui features"() {
+        when:
+        Features features = getFeatures(['swagger-ui'])
+
+        then:
+        features.contains("openapi")
+    }
+
+    void "test config without security feature"() {
+        when:
+        GeneratorContext ctx = buildGeneratorContext(['swagger-ui'])
+        def output = generate(ApplicationType.DEFAULT, ctx)
+
+        then:
+        ctx.configuration.get('micronaut.router.static-resources.swagger.paths') == "classpath:META-INF/swagger"
+        ctx.configuration.get('micronaut.router.static-resources.swagger.mapping') == "/swagger/**"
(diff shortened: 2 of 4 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/other/template/openApiProperties.rocker.raw
```diff
@@ -0,0 +1,6 @@
+swagger-ui.enabled=true
+redoc.enabled=true
+rapidoc.enabled=true
```
graemerocher: Perhaps these should be false if they are not exposed?

n0tl3ss: yea we could also delete those and leave only
```
swagger-ui.enabled=true
```

What do you think is better?

graemerocher: perhaps leave them there set to false so that users know they are available

starter-core/src/main/java/io/micronaut/starter/feature/other/template/openApiProperties.rocker.raw
```diff
@@ -0,0 +1,6 @@
+swagger-ui.enabled=true
+redoc.enabled=false
+rapidoc.enabled=false
+rapidoc.bg-color=#14191f
+rapidoc.text-color=#aec2e0
+rapidoc.sort-endpoints-by=method
```
ilopmar, line 6: I'm not sure if this is the best approach for everyone. There are other options to enable this: using system properties, configuring it in Gradle/Maven,... https://micronaut-projects.github.io/micronaut-openapi/latest/guide/index.html#openApiViews 
I don't know if all users would like this.

Also, if Micronaut Security is present the mappings probably won't work,

graemerocher: I don't think it is worth complicating the PR with different configuration for the different build tools personally

n0tl3ss: If security module is present I can add next part to config:
```
  security:
    intercept-url-map:
      - pattern: /swagger/**
        access:
          - isAnonymous()
      - pattern: /swagger-ui/**
        access:
          - isAnonymous()
```

I guess without this you need to be authenticated to be able to see the swagger-ui?

ilopmar: Be careful with that because maybe not all users want the openapi to be public. Maybe it is safest to add it but commented out and with a comment explaining it

n0tl3ss: I have problem only with generating this part commented out because it is part of application.yaml config file. I am not sure does the Micronaut support separated config files. If it does I can create file with name "application-security.yaml" for an example but I think that the name "security" will resemble environment in this case. Separated file I could generate using the RockerTemplate instead of generating it with the snakeyaml serialization.

pgressa: personally I would keep the swagger ui publicly accessible in case the micronaut-security feature is present in context. I think that way the user can always start the app and immediately test the feature.

What would also be nice is to add annotation for the example controller so there would actually something present in the swagger-ui.

pgressa (changes requested):
add tests

starter-core/src/test/groovy/io/micronaut/starter/feature/other/SwaggerUISpec.groovy
```diff
@@ -0,0 +1,74 @@
...
+        when:
+        GeneratorContext ctx = buildGeneratorContext(['swagger-ui', 'security'])
+        def output = generate(ApplicationType.DEFAULT, ctx)
+
+        then:
+        ctx.configuration.get('micronaut.router.static-resources.swagger.paths') == "classpath:META-INF/swagger"
+        ctx.configuration.get('micronaut.router.static-resources.swagger.mapping') == "/swagger/**"
+        ctx.configuration.get('micronaut.router.static-resources.swagger-ui.paths') == "classpath:META-INF/swagger/views/swagger-ui"
+        ctx.configuration.get('micronaut.router.static-resources.swagger-ui.mapping') == "/swagger-ui/**"
+        ctx.configuration.get('micronaut.router.static-resources.swagger-ui.mapping') == "/swagger-ui/**"
```
pgressa: duplicated line

test-core/src/test/groovy/io/micronaut/starter/core/test/feature/other/SwaggerUISpec.groovy
```diff
@@ -0,0 +1,194 @@
...
+        checkOpenApiPropertiesFile()
+
+        checkControllerFile(language)
+
+        checkYAMLConfigFileWithoutSecurityFeature()
+
+        where:
+        [feature, language] << [
+                beanContext.getBeansOfType(SwaggerUI),
+                [Language.JAVA, Language.KOTLIN]].combinations()
```
pgressa: why not groovy?

pgressa (changes requested):
just duplicated line, but otherwise I think good to go. Good job!

pgressa (approved):

starter-core/src/main/java/io/micronaut/starter/feature/other/template/openApiProperties.rocker.raw
scprek: @n0tl3ss thanks for doing this, I was looking for this feature as well. One question, you have redoc/rapidoc in the properties file, but you are not generating the static resource in the application.conf file. Are you planning on adding all possible? Seems like it should at least be consistent between the config and properties file?

```yaml
  router:
    static-resources:
      swagger:
        paths: classpath:META-INF/swagger
        mapping: /swagger/**
      swagger-ui:
        paths: classpath:META-INF/swagger/views/swagger-ui
        mapping: /swagger-ui/**
      redoc:
        paths: classpath:META-INF/swagger/views/redoc
        mapping: /redoc/**
```

n0tl3ss: @graemerocher should I add and enable redoc as well?

n0tl3ss: @scprek With these code changes, redoc view won't be generated. In `openapi.properties` as suggested above I changed `redoc.enabled` to `false`. There are only two routes enabled, one points to all the swagger resources `/swagger` and one that makes a shortcut to swagger-ui `/swagger-ui`  which points to `/swagger/views/swagger-ui` folder. All those files/folders are in `META-INF` folder.

scprek: You are correct. I guess I meant to ask if you should enable all of them and then a user can decide which to disable.

n0tl3ss: I would leave it like this. Even if we enable redoc by default I wouldn't add the shorcut route `/redoc` to `application.yml` conifg file because when someone want to disable it and sets it to `false` there is potential that `/redoc` will be left as dangling route. If redoc is enabled you could access it via `/swagger/views/redoc` route.  It shouldn't be hard to add. If others agrees to add it I don't mind.

scprek: Yeah that's a good point. I think this is fine then.

PR #1059, opened 2022-01-31
remove SnakeCaseStrategy as it is surplus

### LINKED ISSUE #999: "Push to GitHub" fails when using 2FA
If the GitHub account has 2FA active, after clicking "Push to GitHub" and authorizing the OAuth request, you get the error `Failed to get user access token.`

If it's not practical to request and send a 2FA token, we should detect that 2FA is active and/or display a more informative error message.

n0tl3ss:
#999 push to github doesn't work even without 2FA. Here is a stack trace:

```
ERROR i.m.r.intercept.RecoveryInterceptor - Type [io.micronaut.starter.client.github.oauth.GitHubOAuthClient$Intercepted] executed with error: Error decoding HTTP response body: Error decoding stream for type [class io.micronaut.starter.client.github.oauth.AccessToken]: Class com.fasterxml.jackson.databind.PropertyNamingStrategies$SnakeCaseStrategy has no default (no arg) constructor
```

Removing ```@JsonNaming``` should fix the problem.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/client/github/oauth/AccessToken.java
+++ b/starter-core/src/main/java/io/micronaut/starter/client/github/oauth/AccessToken.java
@@ -17,8 +17,6 @@ package io.micronaut.starter.client.github.oauth;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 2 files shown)


pgressa:
so basically revert of https://github.com/micronaut-projects/micronaut-starter/commit/0cc94c77896bfee135dd3009a70c9995f01183fd

pgressa (approved):

PR #1087, opened 2022-03-09
Upgrade to the Gradle plugins 3.3.0

melix:


starter-core/src/test/groovy/io/micronaut/starter/feature/build/MicronautAotBuildPluginSpec.groovy
```diff
@@ -13,7 +13,7 @@ import static io.micronaut.starter.options.BuildTool.GRADLE_KOTLIN
 
 class MicronautAotBuildPluginSpec extends ApplicationContextSpec implements CommandOutputFixture {
 
-    private static final String GRADLE_PLUGIN_VERSION = '3.2.2'
+    private static final String GRADLE_PLUGIN_VERSION = '3.3.0'
```
sdelamo, line 16: this should not have the plugin version harcoded, instead it should retrieve it from the pom.xml

sdelamo (changes requested):

starter-core/src/test/groovy/io/micronaut/starter/feature/build/MicronautAotBuildPluginSpec.groovy
sdelamo: It is in a test. nevermind.

PR #1123, opened 2022-03-21
Parse the XML when testing the Maven Gradle Enterprise plugin

timyates:
- ~~See if we gain any speed boost~~ -- we don't
- Fix spock test to parse the XML


timyates:
Result:  No difference...

Pre `maxParallelForks = 2` 22m : https://ge.micronaut.io/s/p6uarrgu4mnto/timeline 
With `maxParallelForks = 2` 25m : https://ge.micronaut.io/s/druw6pjeqtt3e/timeline

sdelamo (approved):

PR #1159, opened 2022-04-05
Add the Gradle Plugin portal as a repository for dependabot

timyates:
Dependabot doesn't see

```
        <dependency>
            <groupId>com.gradle</groupId>
            <artifactId>gradle-enterprise-gradle-plugin</artifactId>
            <version>3.8.1</version>
        </dependency>
```

(from our POM) In maven central, so doesn't suggest updates to the plugin version to keep us up to date.

This change ([suggested in slack](https://github.com/gradle/wrapper-upgrade-gradle-plugin/blob/main/.github/dependabot.yml#L3-L14)) adds the Gradle Plugin Portal as a known repository for dependabot, so it should keep us up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,17 @@
 version: 2
+registries:
+  gradle-plugin-portal:
+    type: maven-repository
+    url: https://plugins.gradle.org/m2
+    username: dummy # Required by dependabot
+    password: dummy # Required by dependabot
 updates:
+  - package-ecosystem: "gradle-plugins"
+    directory: "/"
+    registries:
+      - gradle-plugin-portal
+    schedule:
+      interval: "daily"
   - package-ecosystem: "gradle"
     directory: "/"
     schedule:


alvarosanchez:
Doesn't it need to be changed here too? https://github.com/micronaut-projects/micronaut-project-template/blob/master/.github/dependabot.yml

timyates:
Do we use gradle plugins that require updating in this manner in any other projects?

sdelamo:
> Do we use gradle plugins that require updating in this manner in any other projects?

Yes, There are places where we use gradle plugins in modules. 

For example: https://github.com/micronaut-projects/micronaut-views/blob/master/views-jte/build.gradle#L3

sdelamo (approved):

timyates:
> Yes, There are places where we use gradle plugins in modules.

But does dependabot see those? I thought this would just fix it for when it's in the POM?

sdelamo:
> > Yes, There are places where we use gradle plugins in modules.
> 
> But does dependabot see those? I thought this would just fix it for when it's in the POM?

No idea 🤷🏼‍♂️. 

Probably not. I will be surprised if dependabot is able to submit PR to increase a Gradle plugin version.

PR #1162, opened 2022-04-06
As of 3.5.0, GRPC is now a micronaut bom property

timyates:
So it has changed in `micronaut-versions.properties` from 'grpc' to 'micronaut.grpc'


timyates:
Currently failing for Maven, and I'm not sure why... @alvarosanchez any ideas? (as our resident maven expert) 😉

timyates:
Think I found it... `com.github.os72:protoc-jar-maven-plugin:3.11.4` expects `grpc.version` to be set...  Which it used to be, but now it's not...

Working on a fix

timyates:
I think this needs fixing in micronaut core, not here

PR #1167, opened 2022-04-07
AOT improvements

alvarosanchez:
This PR:

* Removes the hardcoded version in the Gradle file (since it's optional).
* Regards the AOT feature as preview.
* Updates the generated pom.xml when AOT is selected.
* Generates AOT configuration files for Maven.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/MicronautAot.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/MicronautAot.java
@@ -72,11 +72,6 @@ public class MicronautAot implements Feature {
     }
 
     @Override
-    public boolean isPreview() {
-        return true;
-    }
-
-    @Override
     public void apply(GeneratorContext generatorContext) {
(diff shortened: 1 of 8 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/build/MicronautAot.java
```diff
@@ -65,17 +71,28 @@ public boolean supports(ApplicationType applicationType) {
         return applicationType == ApplicationType.DEFAULT;
     }
 
+    @Override
+    public boolean isPreview() {
+        return true;
```
melix: This isn't a preview anymore since we released 1.0.0.

melix (approved):

starter-core/src/main/java/io/micronaut/starter/feature/build/MicronautAot.java
alvarosanchez: I've added a comment in the build files instead.

PR #1176, opened 2022-04-11
Allow fetching version props from included boms

### LINKED ISSUE #1173: grpc.version is required in the properties to build GRPC apps
In 3.4.x (pre-grpc bom), micronaut-core contained a managed version for grpc

https://github.com/micronaut-projects/micronaut-core/blob/0b91dbfc8410e70aa90e9aa670be959fad6167bf/gradle/libs.versions.toml#L44

This was then part of the [3.4.x bom](https://search.maven.org/artifact/io.micronaut/micronaut-bom/3.4.1/pom)

```
    <grpc.version>1.39.0</grpc.version>
```

When we build starter, we parse the BOM, and extract all of the properties out into a properties file [in this task](https://github.com/micronaut-projects/micronaut-starter/blob/3.4.x/buildSrc/src/main/groovy/io/micronaut/internal/starter/tasks/WriteMicronautVersionInfoTask.java#L66-L91)

So in 3.4.x, we get a file `starter-core/build/version-info/micronaut-versions.properties` which contains

```
grpc.version=1.39.0
```

This is used by our Gradle project builder to define the GRPC plugin version for the protobuf builder

For the Kotlin DSL:
https://github.com/micronaut-projects/micronaut-starter/blob/ec8d30a7521678d0341542aae3f6f869f32fa82c/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw#L236-L240

And the Groovy DSL:
https://github.com/micronaut-projects/micronaut-starter/blob/ec8d30a7521678d0341542aae3f6f869f32fa82c/starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw#L253-L255

It is also used (silently) by the Maven `protoc-jar-maven-plugin` plugin the failure if this property isn't set [can be seen here](https://github.com/micronaut-projects/micronaut-starter/runs/5849279038?check_suite_focus=true)

## Solutions!

There are 2 solutions I can think of, and both of them don't taste 100% delicious.

I am also not sure if this is just an issue with the way we build starter (as I suspect), or whether it would be something that affects the wider community...

### 1.  Add the GRPC managed version back into the micronaut-core BOM

That is what I have done here with this PR: https://github.com/micronaut-projects/micronaut-core/pull/7190

#### The Good

This works.

#### The bad.

We have to keep the 2 in sync...  If we update the grpc version in the module, we MUST ensure the version is updated in the bom

### 2.  Change the property extraction task to pull out versions from boms defined in the parent core bom

This could be a global "extract all the versions from the entire micronaut bom tree", or a limited "we know we need grpc, so get the versions for just that bom as well"

#### The Good

This stops us having to keep things in sync.

#### The bad

We may have version collisions.  However, this could be minimized by limiting the boms we traverse to just those we know we need for the build.
We cannot prefix the version names with something to avoid collisions, as the Maven plugin requires __exactly__ `grpc.version`

timyates:
We require the grpc.version from the grpc bom.  This is no longer managed by the core bom.

This change allows setting a list of projects to fetch the bom form, and these properties are
added to those extracted from the core bom

Fixes #1173

I believe we would need to do similar in core

### DIFF
--- a/buildSrc/src/main/groovy/io/micronaut/internal/starter/tasks/WriteMicronautVersionInfoTask.java
+++ b/buildSrc/src/main/groovy/io/micronaut/internal/starter/tasks/WriteMicronautVersionInfoTask.java
@@ -23,6 +23,7 @@ import org.gradle.api.artifacts.result.ArtifactResolutionResult;
 import org.gradle.api.artifacts.result.ComponentArtifactsResult;
 import org.gradle.api.artifacts.result.ResolvedArtifactResult;
 import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.Property;
 import org.gradle.api.tasks.CacheableTask;
 import org.gradle.api.tasks.Input;
@@ -45,6 +46,8 @@ public abstract class WriteMicronautVersionInfoTask extends DefaultTask {
     @Input
     public abstract Property<String> getVersion();
 
+    @Input abstract ListProperty<String> getExtraBomProperties();
+
     @OutputDirectory
     public abstract DirectoryProperty getOutputDirectory();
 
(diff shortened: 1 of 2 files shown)


alvarosanchez (approved):

timyates:
Maven will fail until https://github.com/micronaut-projects/micronaut-core/pull/7202 I believe

melix (approved):

sdelamo (approved):

PR #1181, opened 2022-04-13
Add spring-boot-annotation as annotationProcessor dependency

mattmoss:
Add spring-boot-annotation as annotationProcessor dependency during project generation with spring-boot feature.

When using starter to generate a project with the spring-boot feature, it adds the dependency:

`
    annotationProcessor("io.micronaut.spring:micronaut-spring-annotation")
`

(Also same for `testAnnotationProcessor`.) However, what is needed is `micronaut-spring-boot-annotation`, in order to have the following annotation mappers included:

[https://github.com/micronaut-projects/micronaut-spring/blob/master/spring-boot-annotation/src/main/resources/META-INF/services/io.micronaut.inject.annotation.AnnotationMapper]()


sdelamo (changes requested):
I this is correct, we need to change the Micronaut spring documentation: 


https://micronaut-projects.github.io/micronaut-spring/snapshot/guide/index.html#springBoot

sdelamo (approved):

PR #1186, opened 2022-04-25
Add a new command to create an application builder style

jameskleeh:
render1650906677632

This is a little outdated now because I added a blue `> ` prompt where users enter the selection

### DIFF
--- a/starter-cli/src/main/java/io/micronaut/starter/cli/MicronautStarter.java
+++ b/starter-cli/src/main/java/io/micronaut/starter/cli/MicronautStarter.java
@@ -47,7 +47,8 @@ import java.util.function.BiFunction;
                 CreateCliCommand.class,
                 CreateFunctionCommand.class,
                 CreateGrpcCommand.class,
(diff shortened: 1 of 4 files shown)


graemerocher (approved):

sdelamo (commented):
Tab autocomplete does not work for me. Am I testing it wrong?

AutocompleteBroken

sdelamo:
It seems if I execute `java -jar starter-cli/build/libs/micronaut-cli-3.5.0-SNAPSHOT-all.jar create`. Tab automcompletion works. It works also for native image.

sdelamo (approved):

sdelamo:
I have tested with native image as well. It work well with MacOS. Not sure if the Ansi logic will give any issues with Windows.

sdelamo:
Scott De Vaughn tested it in in Windows as well. It works both the native and jar. 

Thanks @jameskleeh
For me this is going to be a time saver.

PR #1189, opened 2022-04-27
Fixing issues with the micronaut-serialization features

### LINKED ISSUE #1166: Include dependency substitution block for serialization modules
### Description

When selecting a `micronaut-serialization` feature (jackson, bson, ...) there's a dependency substitution block which needs to be set (see https://micronaut-projects.github.io/micronaut-serialization/1.0.x/guide/#jacksonQuick).

If the substitution is not added, the application would start but would contain more dependencies than needed and possibly not using the serialization module at all (hard to tell because the runtime behavior is correct).

wetted:
- Adding dependency substitution blocks for micronaut-serialization features for gradle builds, as well as the necessary runtime exclusions for maven builds.

- Add `jakarta.json.bind-api` artifact to pom.xml, for version lookup applied to gradle dependency substitution blocks

Fixes #1166

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/build/BuildPlugin.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/BuildPlugin.java
@@ -22,15 +22,13 @@ import io.micronaut.starter.build.dependencies.CoordinateResolver;
 import io.micronaut.starter.options.BuildTool;
 import io.micronaut.starter.template.Writable;
 
-import java.util.List;
-
 public interface BuildPlugin extends Ordered {
 
     @NonNull
     BuildTool getBuildTool();
 
     @Nullable
-    List<Writable> getExtensions();
+    Writable getExtension();
 
     boolean requiresLookup();
 
--- a/starter-core/src/main/java/io/micronaut/starter/build/BuildPlugin.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/BuildPlugin.java
@@ -22,13 +22,15 @@ import io.micronaut.starter.build.dependencies.CoordinateResolver;
 import io.micronaut.starter.options.BuildTool;
 import io.micronaut.starter.template.Writable;
 
+import java.util.List;
+
 public interface BuildPlugin extends Ordered {
 
     @NonNull
     BuildTool getBuildTool();
 
     @Nullable
-    Writable getExtension();
+    List<Writable> getExtensions();
 
     boolean requiresLookup();
 
--- a/starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
@@ -16,10 +16,14 @@
 package io.micronaut.starter.build.gradle;
 
 import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.util.CollectionUtils;
+import io.micronaut.starter.build.dependencies.DependencyCoordinate;
+import io.micronaut.starter.build.dependencies.Substitution;
+import io.micronaut.starter.template.RockerWritable;
 import io.micronaut.starter.template.Writable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import io.micronaut.starter.feature.build.gradle.templates.substitutions;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
@@ -74,6 +78,19 @@ public class GradleBuild {
     @NonNull
     public List<GradlePlugin> getSettingsPlugins() {
         return plugins.stream().filter(gradlePlugin -> gradlePlugin.getGradleFile() == GradleFile.SETTINGS).collect(Collectors.toList());
+    }
+
+    @NonNull
+    public String renderSubstitutions() {
+        Set<Substitution> uniqueSubstitutions = new HashSet<>();
+        dependencies
+                .stream()
+                .map(DependencyCoordinate::getSubstitutions)
+                .filter(Objects::nonNull)
+                .forEach(uniqueSubstitutions::addAll);
+        return CollectionUtils.isEmpty(uniqueSubstitutions) ? "" :
+                renderWritableExtensions(Stream.of(
+                        new RockerWritable(substitutions.template(uniqueSubstitutions))));
     }
 
     @NonNull
--- a/starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
@@ -23,6 +23,7 @@ import org.slf4j.LoggerFactory;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -77,7 +78,12 @@ public class GradleBuild {
 
     @NonNull
     public String renderExtensions() {
-        return renderWritableExtensions(plugins.stream().map(GradlePlugin::getExtension));
+        List<Writable> extensions = new ArrayList<>();
+        plugins.stream()
+                .map(GradlePlugin::getExtensions)
+                .filter(Objects::nonNull)
+                .forEach(extensions::addAll);
+        return renderWritableExtensions(extensions.stream());
     }
 
     @NonNull
--- a/starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
@@ -27,7 +27,6 @@ import io.micronaut.starter.feature.build.gradle.templates.substitutions;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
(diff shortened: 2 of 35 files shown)


wetted:
There are code smells in this PR because `micronaut-runtime` artifact is included automatically by the `lang` features. This is problematic because for the `maven` build versions of the feature, we need to add an exclusion to a dependency that's already been applied. If we have more examples of dependencies that need substitutions, exclusions or other qualifications we might need an API to support it. I also added a delegate method to `Features` since the `GeneratorContext` isn't exposed to the rocker template (although it could be), and I needed to lookup a pom artifact version. An API that solves the first problem might make that unnecessary.

sdelamo:
@wetted I have added capability for adding subsitutions and exclusions.  Check it out and let me know if you have any concerns.

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/substitutions.rocker.raw
```diff
@@ -0,0 +1,12 @@
+@import io.micronaut.starter.build.dependencies.Substitution
+@import java.util.Set
+@args (Set<Substitution> substitutions)
+
+configurations.all {
+    resolutionStrategy.dependencySubstitution {
+@for (substitution : substitutions) {
```
alvarosanchez, line 8: What if `substitutions` is null or empty?

starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
```diff
@@ -16,13 +16,18 @@
...
 import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.util.CollectionUtils;
+import io.micronaut.starter.build.dependencies.DependencyCoordinate;
+import io.micronaut.starter.build.dependencies.Substitution;
+import io.micronaut.starter.template.RockerWritable;
 import io.micronaut.starter.template.Writable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import io.micronaut.starter.feature.build.gradle.templates.substitutions;
```
yawkat, line 26: what is this?

starter-core/src/main/java/io/micronaut/starter/build/BuildPlugin.java
```diff
@@ -22,13 +22,15 @@
...
+import java.util.List;
+
 public interface BuildPlugin extends Ordered {
 
     @NonNull
     BuildTool getBuildTool();
 
     @Nullable
-    Writable getExtension();
+    List<Writable> getExtensions();
```
yawkat: maybe it's easier to just combine the multiple extensions into a single `Writable` instance, could reduce the required changes

wetted:
@sdelamo  I believe the substitution for `SerializationBsonFeature` isn't correct. If you add the `Book` example from the `micronaut-serialization` docs to a project created by starter, it won't compile, with an error importing `jakarta.json.bind.annotation.*`. I think it needs to be consistent with `SerializationJsonpFeature` feature where the substitution adds `jakarta.json.bind-api`.

However, @graemerocher raised a question about that when I corrected the docs.  Please see https://github.com/micronaut-projects/micronaut-serialization/pull/195

starter-core/src/main/java/io/micronaut/starter/feature/json/SerializationBsonFeature.java
```diff
@@ -38,4 +48,31 @@ public String getTitle() {
     public String getModule() {
         return "bson";
     }
+
+    @Override
+    @NonNull
+    public List<Substitution> substitutions(@NonNull GeneratorContext generatorContext) {
```
yawkat, line 54: bson cant really substitute json, not sure if it really makes sense to exclude jackson here (though youd probably also be using serde json if youre using this feature)

starter-core/src/main/java/io/micronaut/starter/feature/json/SerializationFeature.java
```diff
@@ -15,12 +15,22 @@
...
 import io.micronaut.starter.feature.Category;
+import io.micronaut.starter.options.BuildTool;
+import java.util.ArrayList;
+import java.util.List;
 
 public interface SerializationFeature extends JsonFeature {
+    String MICRONAUT_SERIALIZATION = "micronaut.serialization";
+    String GROUP_ID_MICRONAUT = "io.micronaut";
+    String GROUP_ID_MICRONAUT_SERDE = "io.micronaut.serde";
+    String ARTIFACT_ID_MICRONAUT_JACKSON_DATABIND = "micronaut-jackson-databind";
```
yawkat: might as well make constants for the full Dependency object

alvarosanchez (approved):
Great job with this PR @wetted and @sdelamo !

starter-core/src/main/java/io/micronaut/starter/feature/json/SerializationFeature.java
```diff
@@ -44,20 +54,54 @@ default boolean supports(ApplicationType applicationType) {
...
+
+    @NonNull
+    default Dependency.Builder micronautRuntimeDependency() {
+        return Dependency.builder()
+                .compile()
+                .groupId(GROUP_ID_MICRONAUT)
+                .artifactId("micronaut-runtime")
+                .exclude(Dependency.builder()
+                        .groupId("io.micronaut")
+                        .artifactId("micronaut-jackson-databind")
```
yawkat: use constant

starter-core/src/main/java/io/micronaut/starter/feature/json/SerializationFeature.java
yawkat: Something like:

```java
Dependency MICRONAUT_JACKSON_DATABIND = Dependency.builder()
                                .groupId(GROUP_ID_MICRONAUT)
                                .artifactId(ARTIFACT_ID_MICRONAUT_JACKSON_DATABIND)
                                .build();
```

starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
timyates: That's the class that rocker generates from the `substitutions.rocker.raw` file

timyates (approved):
@melix can you check this fixes the issue you had?

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/substitutions.rocker.raw
sdelamo: we never call this template but to be safe I can wrap the whole thing in an if statement.

starter-core/src/main/java/io/micronaut/starter/feature/json/SerializationBsonFeature.java
sdelamo: I think the substitutions here match the one in the docs: 

https://micronaut-projects.github.io/micronaut-serialization/latest/guide/#bsonQuick

Are the subsitutions in the docs incorrect?

starter-core/src/main/java/io/micronaut/starter/feature/json/SerializationFeature.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-starter/pull/1189/commits/877f9c79e195799c66012e9f36c5b69831f5cb9c

starter-core/src/main/java/io/micronaut/starter/feature/json/SerializationBsonFeature.java
wetted: @sdelamo see my comment about that above. I have a PR to change the docs, but that may still be different from what @yawkat is says.

starter-core/src/main/java/io/micronaut/starter/feature/json/SerializationFeature.java
sdelamo: I extracted it since every JsonFeature uses it: 

https://github.com/micronaut-projects/micronaut-starter/pull/1189/commits/6826a62e77a6b48845b6f981e678f39aaa8dddea

starter-core/src/main/java/io/micronaut/starter/build/BuildPlugin.java
sdelamo: I reverted this: 

https://github.com/micronaut-projects/micronaut-starter/pull/1189/commits/6421e0ef0141f3010069946dbeef148617174c62

starter-core/src/main/java/io/micronaut/starter/feature/json/SerializationBsonFeature.java
sdelamo: I think we have to fix the docs to be correct and then update this PR to match what the docs say

wetted: I have a PR for that at https://github.com/micronaut-projects/micronaut-serialization/pull/195

yawkat: Ah yea, i forgot that the bson module can just use the bson library for serialization of json, still.

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/substitutions.rocker.raw
sdelamo: I have wrapped it in an if statement

PR #1194, opened 2022-05-02
Upgrade to Gradle plugin 3.4.0

melix:
This commit bumps the Gradle plugin version to 3.4.0. It takes advantage of this to bump the Micronaut AOT version to 1.1.0.

Also bumps infrastructure (build plugins) to latest.


alvarosanchez:
Enabling AOT optimizations via a Gradle DSL introduces a maintenance burden and increases the gap between Maven and Gradle.

If you guys are moving forward with this way, can you at least create an issue to perform the same change for Maven builds, please?

PR #1205, opened 2022-05-04
Add MicroStream features

### LINKED ISSUE #1187: Create a microstream feature
It will need to add: 

https://github.com/micronaut-projects/micronaut-microstream/blob/master/test-suite/build.gradle#L8-L9

We need to release the module or at least a RC before releasing this feature. 

We may need several features, one for core, another for cache and another for rest

timyates:
One for the main module, one for the REST API and another for caching

Fixes #1187


starter-core/src/main/java/io/micronaut/starter/feature/microstream/MicroStream.java
```diff
@@ -0,0 +1,93 @@
...
+        generatorContext.addDependency(Dependency.builder()
+                .compile()
+                .groupId(MICRONAUT_MICROSTREAM_GROUP_ID)
+                .artifactId("micronaut-microstream")
+                .build()
+        );
+        generatorContext.addDependency(Dependency.builder()
+                .compile()
+                .groupId(MICRONAUT_MICROSTREAM_GROUP_ID)
+                .artifactId("micronaut-microstream-annotations")
```
sdelamo, line 74: we need this dependency as an annotation processor as well.

starter-core/src/main/java/io/micronaut/starter/feature/microstream/MicroStreamCache.java
```diff
@@ -0,0 +1,93 @@
...
+        return Category.CACHE;
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.getConfiguration().put("microstream.storage.cachestore.storage-directory", "/tmp/cache");
+        generatorContext.getConfiguration().put("microstream.cache.my-cache.key-type", "java.lang.Integer");
+        generatorContext.getConfiguration().put("microstream.cache.my-cache.value-type", "java.lang.String");
+        generatorContext.getConfiguration().put("microstream.cache.my-cache.backing-storage", "cachestore");
+        generatorContext.addDependency(Dependency.builder()
```
sdelamo: this is no longer necessary, we expose it transitively.

starter-core/src/main/java/io/micronaut/starter/feature/microstream/MicroStreamRest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Override
+    public String getCategory() {
+        return Category.DATABASE;
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.getConfiguration().put("microstream.rest.enabled", "true");
+        generatorContext.getConfiguration().put("microstream.rest.path", "microstream");
```
sdelamo: I don't think we should set the path. The default should be enough.

starter-core/src/main/java/io/micronaut/starter/feature/microstream/MicroStreamRest.java
```diff
@@ -0,0 +1,99 @@
...
+        generatorContext.addDependency(Dependency.builder()
```
sdelamo, line 72: for gradle this dependency should be `developmentOnly`. we don't want to expose the database contents accidentally.

starter-core/src/main/java/io/micronaut/starter/feature/microstream/MicroStreamRest.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public boolean supports(ApplicationType applicationType) {
+        return true;
+    }
+
+    @Override
+    public String getCategory() {
+        return Category.DATABASE;
```
sdelamo: DEV_TOOLS maybe?

starter-core/src/main/java/io/micronaut/starter/feature/microstream/MicroStream.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public String getCategory() {
+        return Category.DATABASE;
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.getConfiguration().put("microstream.storage.main.storage-directory", "/tmp/microstream");
```
sdelamo: I don't we should add configuration entries. There is really not sensible defaults. Windows users may not have a `/tmp` folder. What do you think if we removed both lines: 

```
        generatorContext.getConfiguration().put("microstream.storage.main.storage-directory", "/tmp/microstream");
Write
        generatorContext.getConfiguration().put("microstream.storage.main.root-class", "java.util.ArrayList");
```

starter-core/src/main/java/io/micronaut/starter/feature/microstream/MicroStreamCache.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public String getCategory() {
+        return Category.CACHE;
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.getConfiguration().put("microstream.storage.cachestore.storage-directory", "/tmp/cache");
```
sdelamo: I will remove these configuration entries.

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/microstream/MicroStreamRest.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.getConfiguration().put("microstream.rest.enabled", "true");
```
sdelamo: don't enable it by default. User should have to opt-in. I don't want people to inadvertently expose their contents of their storage.

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/microstream/MicroStreamRest.java
timyates: We don't have a way to specify developmentOnly that I can find...  Do you want to add it?

sdelamo: We do now 😎

https://github.com/micronaut-projects/micronaut-starter/pull/1205/commits/c440999e01618871f0b9b235353627fc5307da3e

PR #1207, opened 2022-05-06
Micronaut Console Feature

musketyr:
Add [Micronaut Console](https://agorapulse.github.io/micronaut-console/) to the development tools features.

The feature adds dependency on `com.agorapulse:micronaut-console` regardless the language. For Java projects, it also adds `org.codehaus.groovy:groovy` dependency to enable Groovy support for the console. For Kotlin projects, it also adds `org.jetbrains.kotlin:kotlin-scripting-jsr223` to enable Kotlin scripting. Example IntelliJ HTTP file is generated into `src/test/resources` for every project. For Java and Groovy projects, also the GDSL file is generated to enable content assist. The `src/test/resources` has been selected because it enables the content assist in IntelliJ but the files are not part of the production files.


graemerocher:
Since this allows running arbitrary code is it locked down by default?

musketyr:
@graemerocher see the [Security section](https://agorapulse.github.io/micronaut-console/#_security). The console is disabled by default for `cloud` environment which should cover all the production use cases. Let me know if there know some other environments where it should be disabled by default.

musketyr:
If there is a way how to contribute the `application.yml` file we can consider adding the [forgery protection](https://agorapulse.github.io/micronaut-console/#_server_side_request_forgery_protection) by default.

alvarosanchez:
I'd probably set `console.enabled = false` by default and instruct users to explicitly opt-in for whatever environments/configurations they want

alvarosanchez:
BTW you can contribute configuration via [`GeneratorContext`](https://github.com/micronaut-projects/micronaut-starter/blob/3.4.x/starter-core/src/main/java/io/micronaut/starter/application/generator/GeneratorContext.java#L152)

musketyr:
> I'd probably set `console.enabled = false` by default and instruct users to explicitly opt-in for whatever environments/configurations they want

I think it would be nice if the use can try the console right ahead in the local environment. I can add the SSRF token protection + localhost only by default. This is how we secure our production endpoints.

alvarosanchez:
> I can add the SSRF token protection + localhost only by default

I like that better than simply considering the `cloud` env. Not all users live in the cloud, unfortunately.

musketyr:
👍 @alvarosanchez 

see https://github.com/micronaut-projects/micronaut-starter/pull/1207/commits/33c9a52840b19cd3e8862f179aea10326b8d6eb4

The console is now restricted to localhost and guarded by verification header.
Snímek obrazovky 2022-05-06 v 11 56 23

alvarosanchez:
Thanks! Could you please assert in a test that the configuration gets added?

musketyr:
@alvarosanchez done https://github.com/micronaut-projects/micronaut-starter/pull/1207/commits/21718ff3ee40d241417b4d6deb4964a2ab886890

alvarosanchez (approved):

alvarosanchez:
Adding @sdelamo to the loop for the 3.5 release notes.

graemerocher:
Can make make these changes to the application configuration the default if no configuration is provided? Sorry to ask you to change your library / framework, but I am really conscious of the security risks here.

musketyr:
ok, version 2.x is now disabled by default see https://github.com/micronaut-projects/micronaut-starter/pull/1207/commits/f5ee68a979d27b3136859a9a36ed6ad39cd2c581 and https://github.com/agorapulse/micronaut-console/commit/7f8865df687838d8f2825abfe822a1065ae356b9

starter-core/src/main/java/io/micronaut/starter/feature/agorapulse/console/Console.java
```diff
@@ -0,0 +1,162 @@
...
+        String secret = UUID.randomUUID().toString();
+        addDependency(generatorContext);
+        addExampleCode(generatorContext, secret);
+        addConfiguration(generatorContext, secret);
+    }
+
+    private void addDependency(GeneratorContext generatorContext) {
+        generatorContext.addDependency(Dependency.builder()
+                .lookupArtifactId(ARTIFACT_ID)
+                .runtime());
```
sdelamo: for Gradle, please use `developmentOnly` not `runtimeOnly`. That way, it is not in production applications by default.

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/agorapulse/console/Console.java
musketyr: sounds good but how exactly can I do it?

sdelamo: It should be possible after this PR gets merged. https://github.com/micronaut-projects/micronaut-starter/pull/1215

musketyr: resolved in https://github.com/micronaut-projects/micronaut-starter/pull/1207/commits/e7339697350e512458318de87a355088ad729b42

sdelamo:
Closed via https://github.com/micronaut-projects/micronaut-starter/pull/1232

PR #1208, opened 2022-05-06
Kapt dependencies in Maven seem to require an order

timyates:
A method was renamed in https://github.com/micronaut-projects/micronaut-core/pull/7051

But as some annotation libs pull in older versions of inject-java, Maven fails if they are
earlier in the dependency list.

This was found investigating here https://github.com/micronaut-projects/micronaut-guides/issues/918

I suspect it's also causing the other Kotlin failures in guides.

This also reverts some of the changes here https://github.com/micronaut-projects/micronaut-starter/pull/1146
when we had the same issue with Mongo-Data processing in Maven

I need to test building the guides against a snapshot of this.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/application/generator/GeneratorContext.java
+++ b/starter-core/src/main/java/io/micronaut/starter/application/generator/GeneratorContext.java
@@ -51,7 +51,6 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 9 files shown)


timyates:
Instead of the comparator, we could use

https://github.com/micronaut-projects/micronaut-starter/blob/f400a00ac2d9781cfcea3ceccd9678e0f667681a/starter-core/src/main/java/io/micronaut/starter/feature/other/ProjectLombok.java#L77

But not sure how we keep Priorities aligned without `Ordered.HIGHEST_PRECEDENCE + 1`

Needs some thought

timyates:
Ok, switched to using an enum

sdelamo:
cc @dstepanov

sdelamo (approved):

PR #1214, opened 2022-05-11
dynamodb feature

sdelamo:
it add support for dynamdob feature

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/aws/DynamoDb.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/aws/DynamoDb.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2022 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/aws/DynamoDb.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public DynamoDb(AwsV2Sdk awsV2Sdk) {
+        this.awsV2Sdk = awsV2Sdk;
+    }
+
+    @Override
+    public void processSelectedFeatures(FeatureContext featureContext) {
+        if (!featureContext.isPresent(AwsV2Sdk.class)) {
+            featureContext.addFeature(awsV2Sdk);
+        }
```
alvarosanchez, line 39: Perhaps we could do this in the parent `AwsFeature`?

alvarosanchez (approved):

starter-core/src/main/java/io/micronaut/starter/feature/aws/DynamoDb.java
sdelamo: maybe. Not sure. We can address that later in a different PR.

PR #1215, opened 2022-05-11
feat: add developmentOnly dependencies

sdelamo:
This allows a feature to contribute a development only dependency.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/build/maven/MavenScope.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/maven/MavenScope.java
@@ -19,6 +19,7 @@ import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.order.Ordered;
 import io.micronaut.starter.build.dependencies.Phase;
 import io.micronaut.starter.build.dependencies.Scope;
+import io.micronaut.starter.build.gradle.GradleConfiguration;
 
 import java.util.Optional;
 
@@ -61,8 +62,10 @@ public enum MavenScope implements Ordered {
                 if (scope.getPhases().contains(Phase.COMPILATION)) {
                     return Optional.of(MavenScope.PROVIDED);
                 }
+                if (scope.getPhases().contains(Phase.DEVELOPMENT)) {
+                    return Optional.of(MavenScope.COMPILE);
+                }
                 break;
-
             case TEST:
                 return Optional.of(MavenScope.TEST);
             default:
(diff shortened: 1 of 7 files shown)


starter-core/src/main/java/io/micronaut/starter/build/maven/MavenScope.java
```diff
@@ -61,8 +62,10 @@ public static Optional<MavenScope> of(@NonNull Scope scope) {
                 if (scope.getPhases().contains(Phase.COMPILATION)) {
                     return Optional.of(MavenScope.PROVIDED);
                 }
+                if (scope.getPhases().contains(Phase.DEVELOPMENT)) {
+                    return Optional.of(MavenScope.COMPILE);
```
alvarosanchez: If in the future we want to support this custom scope in the Maven Plugin, this mapping here will have to be removed (or return empty)

alvarosanchez (approved):

starter-core/src/main/java/io/micronaut/starter/build/maven/MavenScope.java
graemerocher: the mapping is incorrect imo. `developmentOnly` should map to `provided` scope in Maven

graemerocher (changes requested):
Should be `MavenScope.PROVIDED`

starter-core/src/main/java/io/micronaut/starter/build/maven/MavenScope.java
alvarosanchez: `provided` is probably closer, but still incorrect, because they still won't be available at "development runtime"

graemerocher: you mean for `mn:run`? Is that because we aren't resolving provided dependencies in the Maven plugin?

`compile` is definitely wrong, it means it will be included in the production JAR

alvarosanchez: The point here is to allow using the Gradle `developmentOnly` custom configuration. If I'm not mistaken, in Gradle those will be available when running the application, but not anywhere else.

That is not possible with Maven, but we could support it in the Maven Plugin with some additional configuration, since we control the classpath when running `mn:run`.

However, until we get there, those "development only" dependencies will result in provided Maven dependencies, which is not quite the same as `developmentOnly` in Gradle.

melix: If they are not used in Maven, then `PROVIDED` it should be. If we use `compile` they would end up in the distribution, which is not wanted (and has the potential to break things with native image since we include native libraries for file watching).

sdelamo: resolved via https://github.com/micronaut-projects/micronaut-starter/pull/1215/commits/a22938647c55da74f3c4b3acbf795feae55f7e09

alvarosanchez (approved):

PR #1229, opened 2022-05-17
Use TestFilters instead of --tests to appease Test Selection

### LINKED ISSUE #1228: Switch from using --tests to using a test filter
> All of the micronaut-starter › :test-core:test tasks still use --tests which forces all tests to be selected.
> It would be better to use https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/testing/TestFilter.html or disable PTS for this task, as you effectively lose the benefits of Build Caching in this scenario.

(From @eriwen the Micronaut/Gradle Slack channel)

timyates:
Fixes #1228

### DIFF
--- a/.github/workflows/gradle-func.yml
+++ b/.github/workflows/gradle-func.yml
@@ -55,7 +55,7 @@ jobs:
           distribution: 'adopt'
           java-version: ${{ matrix.java }}
       - name: Run app Functional Tests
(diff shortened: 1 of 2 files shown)


melix (approved):
:1st_place_medal:

sdelamo (approved):

PR #1240, opened 2022-05-20
Made Micronaut Console runtime dependency

musketyr:
Please, depending on result of this conversation https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/466 make the Micronaut Console working out of the box for `./gradlew run` task.


sdelamo:
I am not comfortable making this feature a runtime dependency.

sdelamo:
you can add instructions to the README file to tell people to use `-t`

musketyr:
I can see your concerns but to be honest the console is barely used locally. The main usage is in production to debug state of the application. The generated security configuration is the same we are using on our servers and it has been battle proofed.

PR #1241, opened 2022-05-20
build: Split tests

timyates:
We run our tests in separate actions.  This is to gain a speedup as multiple tests can be run at the same time.

We used to do this via `test-core:test --tests "..."` in the action command.  After feedback that this interacts poorly with predictive tests, we switched to using multiple test tasks, with a test filter for each.

However, we sometimes see issues with this.

This PR splits the tests into separate modules, and runs the split actions against each module with no filtering of test names.

Pushed to see if CI likes it


sdelamo (approved):

timyates:
### Scans:

I pulled the Java 8 version build scan links out of the action logs, so we can see which tests were run

(as this is the second run, where I fixed a test in features, I'd expect quite a few to not be selected)

| Type | Java | Scan Link |
| --- | --- | --- |
| Aws func test | 8 | https://ge.micronaut.io/s/cjelxuaujaarw |
| Build tool func test | 8 | https://ge.micronaut.io/s/5tcqk7sbwyplk |
| Cli func test | 8 | https://ge.micronaut.io/s/liraa47rm2ysk |
| Core func test | 8 | https://ge.micronaut.io/s/ggr7uxi657ddi |
| Cloud func test | 8 | https://ge.micronaut.io/s/xodxeppyn5dxa |
| Features func test | 8 | https://ge.micronaut.io/s/qywk3bcwp4oru |

sdelamo:
we should have removed functional tests from Java CI

see: https://github.com/micronaut-projects/micronaut-starter/pull/1248/commits/54786d53e4013e319469a652d9a751dbab45e9c5

PR #1296, opened 2022-06-15
Dropping Kotest4 support in starter in favor of Kotest5

### LINKED ISSUE #1216: Consider dropping Kotest4 support in starter and support Kotest5
Next minor version of Micronaut Test supports [Kotest 5](https://github.com/micronaut-projects/micronaut-test/pull/531).

I think we should only support Kotest5 in starter and new applications created in started should use Kotest5 .

wetted:
Closes #1216 

I'm not sure how to handle versioning, etc during the process of migrating from managed Kotest 4 versions exposed by `micronaut-test` and `micronaut-gradle-plugin` so I littered `io.micronaut.starter.feature.test.KoTest` with some comments on what I think needs to happen. Further guidance is appreciated.

I believe we also need to revise `micronaut-gradle-plugin` so that the `MicronautTestRuntime` enum exposes Kotest5 instead of Kotest 4


timyates (commented):
Do we have an Issue to update the plugin to use Kotest5?

Currently, when we perform a Gradle build with

```
    testRuntime("kotest")
```

It injects Kotest4 into the build [as can be seen here](https://ge.micronaut.io/s/c4furpbe5lmhc/dependencies?dependencies=test-kotest&expandAll&focusedDependencyView=versions)

I don't believe this is used, but it is on the classpath...

There's no kotest5 here in the Gradle plugin

https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/374a4ba6f4a1efbe3e8fce73b8e36cb64161de83/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautTestRuntime.java#L39

timyates:
cc @melix ☝️

timyates:
This doesn't happen with the Maven build: https://ge.micronaut.io/s/glfcuxiaw4bzu/dependencies?search=kotest as it doesn't have a `testRuntime` analog.

Indeed, if we add a Kotest5 `MicronautTestRuntime` to the build, we won't need to add the dependencies:

```
io.kotest:kotest-assertions-core-jvm:5.3.1
io.micronaut.test:micronaut-test-kotest5
io.mockk:mockk:1.12.4
io.kotest:kotest-runner-junit5-jvm:5.3.1
```

to Gradle builds, as these would be handled by the plugin

timyates:
I'll get a pr for the plug-in together

timyates:
To move this forward, I think we have 2 options:

1. Accept that we will have both on the classpath for Gradle builds
2. Remove `testRuntime("kotest")` from the Gradle builds for Kotest for now, and add `testRuntime("kotest5")` and remove the dependencies above at a later date (if the PR is accepted)

wetted:
> To move this forward, I think we have 2 options:
> 
> 1. Accept that we will have both on the classpath for Gradle builds
> 2. Remove `testRuntime("kotest")` from the Gradle builds for Kotest for now, and add `testRuntime("kotest5")` and remove the dependencies above at a later date (if the PR is accepted)

Option 2 is the preferable approach.

I will wait until a version of `micronaut-gradle-plugin` with "kotest5" test runtime is released. It should expose all the Gradle dependencies I list in the `KoTest`feature (which in turn is based on the `MicronautTestRuntime.KOTEST_4` enum value). Then I will change the implementation of `KoTest.apply(...)` to add dependencies for Maven only. I'll also modify the gradle templates to use `testRuntime("kotest5")`.

wetted:
Note: Functional tests for projects using kotest will fail until Starter is updated to `micronaut-gradle-plugin` 3.5.0.

timyates (approved):

PR #1300, opened 2022-06-22
Correct Mongo Reactive name and R2DBC preview status

dstepanov:
Should I add a new module and deprecate the async one because of the name change?


sdelamo:
> Should I add a new module and deprecate the async one because of the name change?

I don't think we need to deprecated it. Target 3.6.0 milestone and 3.6.x branch. 

What is the motivation of renaming this from `async` to `reactive`?

dstepanov:
MongoDB driver implementation is reactive, not async.

graemerocher (approved):

sdelamo (approved):

PR #1303, opened 2022-06-25
Update Camunda projects to versions built with Micronaut 3.5.x

tobiasschaefer:

Please merge this PR. It updates all features to versions built with Micronaut 3.5.2

FYI: @sdelamo

### DIFF
--- a/starter-core/src/main/resources/pom.xml
+++ b/starter-core/src/main/resources/pom.xml
@@ -58,24 +58,24 @@
             <version>1.1.0</version>
         </dependency>
         <dependency>
-            <groupId>org.camunda.assert</groupId>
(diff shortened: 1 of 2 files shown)


sdelamo:
@tobiasschaefer could you set the PR base to 3.6.x?

tobiasschaefer:
@sdelamo : thanks for looking into my PR.

I could technically create the PR against 3.6.x but I cannot test it because 3.6.0 hasn't been released (without too much effort).

I'd actually prefer to have this in the next 3.5.x Micronaut Launch release since the current version contained in Micronaut Launch is built with 3.4.x and is not fully compatible with 3.5.x.

So please merge this for 3.5.x and I'll create new releases for 3.6.x when it is available.

PR #1314, opened 2022-07-05
extract methods

sdelamo:


wetted (approved):

starter-core/src/main/java/io/micronaut/starter/feature/ci/workflows/CIWorkflowFeature.java
```diff
@@ -15,11 +15,13 @@
  */
 package io.micronaut.starter.feature.ci.workflows;
 
+import com.fizzed.rocker.RockerModel;
 import io.micronaut.core.annotation.NonNull;
```
wetted, line 19: I think this NonNull is now unused

wetted (changes requested):

PR #1319, opened 2022-07-08
Use url-connection-client for dynamodb and graal

timyates:
The apache client doesn't work with native compilation, so if the user selects
dynamodb and graalvm then we switch to the url-connection one, and exclude the other
from the dependencies

This also fixes exclusions for Gradle which were not previously required

Pushed for 3.5.x incase we do a release of that as mentioned in the team meeting

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleDependency.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleDependency.java
@@ -28,6 +28,8 @@ import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
+import static io.micronaut.core.util.CollectionUtils.isEmpty;
+
 public class GradleDependency extends DependencyCoordinate {
 
     public static final Comparator<GradleDependency> COMPARATOR = (o1, o2) -> {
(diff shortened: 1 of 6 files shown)


starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleDependency.java
```diff
@@ -102,6 +102,18 @@ public String toSnippet() {
         if (isPom() && isKotlinDSL) {
             snippet += ")";
         }
+        if (getExclusions() != null && !getExclusions().isEmpty()) {
```
sdelamo: We have a Micronaut `CollectionsUtils::isEmpty` to simplify this check

starter-core/src/main/java/io/micronaut/starter/feature/aws/DynamoDb.java
```diff
@@ -35,11 +35,29 @@
...
 import io.micronaut.starter.feature.config.ApplicationConfiguration;
 import io.micronaut.starter.feature.config.Configuration;
+import io.micronaut.starter.feature.graalvm.GraalVM;
 import jakarta.inject.Singleton;
 
 @Singleton
 public class DynamoDb implements AwsFeature {
     public static final String ARTIFACTID_DYNAMODB = "dynamodb";
+
+    private static final Dependency.Builder URL_CONNECTION_CLIENT = Dependency.builder()
```
sdelamo: move this statics to AwsSdkV2 feature

sdelamo (commented):
I created a PR

https://github.com/micronaut-projects/micronaut-starter/pull/1320

sdelamo (approved):

PR #1321, opened 2022-07-09
lambda for Java runtime define JAVA_OPTIONS

sdelamo:
see: https://aws.amazon.com/blogs/compute/optimizing-aws-lambda-function-performance-for-java/


sdelamo:
@timyates 🖕

timyates (approved):

PR #1323, opened 2022-07-11
Add Migration datasource if Migration used with HibernateReactive

timyates:
If we use Hibernate Reactive AND a MigrationFeature, we need to specify a datasource and include the
drivers for the chosen database for the migration to be performed

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/database/DatabaseDriverFeature.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/database/DatabaseDriverFeature.java
@@ -70,9 +70,12 @@ public abstract class DatabaseDriverFeature
     private boolean shouldAddJdbcFeature(FeatureContext featureContext) {
         return !featureContext.isPresent(JdbcFeature.class)
                 && !featureContext.isPresent(R2dbcFeature.class)
-                && !featureContext.isPresent(DataHibernateReactive.class)
-                && !featureContext.isPresent(HibernateReactiveJpa.class)
+                && !hasHibernateReactiveWithoutMigration(featureContext)
                 && jdbcFeature != null;
+    }
+
+    private boolean hasHibernateReactiveWithoutMigration(FeatureContext featureContext) {
+        return featureContext.isPresent(HibernateReactiveFeature.class) && !featureContext.isPresent(MigrationFeature.class);
     }
 
     @Override
@@ -114,6 +117,9 @@ public abstract class DatabaseDriverFeature
         }
         if (generatorContext.getFeatures().hasFeature(DataHibernateReactive.class) || generatorContext.getFeatures().hasFeature(HibernateReactiveJpa.class)) {
             getHibernateReactiveJavaClientDependency().ifPresent(dependencies::add);
+            if (generatorContext.isFeaturePresent(MigrationFeature.class)) {
+                getJavaClientDependency().ifPresent(dependencies::add);
+            }
         } else {
             getJavaClientDependency().ifPresent(dependencies::add);
         }
--- a/starter-core/src/main/java/io/micronaut/starter/feature/database/HibernateReactiveFeature.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/database/HibernateReactiveFeature.java
@@ -19,6 +19,8 @@ import io.micronaut.starter.application.generator.GeneratorContext;
 import io.micronaut.starter.feature.Category;
 import io.micronaut.starter.feature.migration.MigrationFeature;
 
+import java.util.Optional;
+
 public abstract class HibernateReactiveFeature extends TestContainersFeature implements JpaFeature {
 
     public static final String JPA_DEFAULT_REACTIVE = "jpa.default.reactive";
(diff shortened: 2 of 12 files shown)


dstepanov:
Have you tried it's working?
I wouldn't map it to `default` data source, it might conflict with reactive services.
Can you create `datasources.migration.password`?

timyates:
> Have you tried it's working?

This is how I had to do it in the Guides 

https://github.com/micronaut-projects/micronaut-guides/pull/1002/files#diff-70be4842b9d61465b2921189ab9d5f80229b5b8688b94f7f022e91cf4389a15a

>  I wouldn't map it to `default` data source, it might conflict with reactive services. Can you create `datasources.migration.password`?

The problem then becomes that Test Resources would see this as two resources, and migration would be performed on the `migration` data source, and JPA against `default`

Which won't work.  I chatted with @melix and I don't think we could see a cleaner option 🤔

dstepanov:
I would assume Test Resources will skip the second datasource when there are values preset

melix:
Test Resources need a way to figure out that the datasources are the "same", e.g spawn a single test container. I have a local spike which uses the _same_ datasource name for both the "migration" and "reactive" datasources (similarly to what we have for r2dbc). If the names are different, then test resources cannot determine if it should reuse an existing container (and more specifically, _what_ container to use). If we don't do this, then the flyway migration would happen in one container, and the actual connection would use a different container.

Alternatively, we could use a different datasource name, say `migration-default`, if there's a `default` hibernate datasource name. This, however, would create a distinction between how we configure R2DBC and Hibernate Reactive, which, arguably, would be confusing for users.

dstepanov:
Yes, but would this prevent creating a second data source:
```
datasources:
  default:
    url: ${jpa.default.properties.hibernate.connection.url}  # <1>
    username: ${jpa.default.properties.hibernate.connection.username} # <2>
    password: ${jpa.default.properties.hibernate.connection.password}
    driverClassName: com.mysql.cj.jdbc.Driver
```

timyates:
> Yes, but would this prevent creating a second data source:

Not 100% sure what you mean @dstepanov 🤔

dstepanov:
I thought that Test Resource would not start a new database if the values for the default datasource were preset.

melix:
So I finally managed to add support for Hibernate Reactive in Test Resources: https://github.com/micronaut-projects/micronaut-test-resources/pull/50

This is a draft PR because it requires a new release of Micronaut Data *and* Micronaut, so that I get the right BOM and avoid hardcoded dependencies. But this spike uses exactly the same technique as with R2DBC, with the same datasource name between `datasources` and `jpa`. So if a datasource with name `default` exists, it is spawned _first_, allowing Flyway to kick in, then reused for `jpa`. I don't have enough knowledge of the internals to understand what could be wrong with it, but my tests are passing. If there are some risks it doesn't work, we should have a test which shows that this is a wrong approach.

timyates:
Would it work with the config as shown here: https://github.com/micronaut-projects/micronaut-guides/pull/1002/files#diff-70be4842b9d61465b2921189ab9d5f80229b5b8688b94f7f022e91cf4389a15a

ie:  How would we configure it so that we only need to supply the url/username and password once when it's put into production?

melix:
Test resources require properties to be _missing_, so it should be like:

```yaml
jpa:
  default:
    entity-scan:
      packages:
        - 'example.micronaut.domain'
    properties:
      hibernate:
        show-sql: true
        hbm2ddl:
          auto: update
    reactive: true
---
datasources:
  default:
    url: ${jpa.default.properties.hibernate.connection.url}
    username: ${jpa.default.properties.hibernate.connection.username}
    password: ${jpa.default.properties.hibernate.connection.password}
    driverClassName: com.mysql.cj.jdbc.Driver
flyway:
  datasources:
    migration:
      enabled: true 
```

melix:
(maybe the other way around is cleaner, that is, the `jpa` configuration should reuse the `datasources.default` one, so that it mimics what test resources will do)

timyates:
I will try and get a manual test running with the above changes and a version from the draft PR 👍

timyates:
Just tested with local snapshots of everything (plugin, TR, etc) and this config

```
jpa:
  default:
    entity-scan:
      packages:
        - 'example.micronaut.domain' # <1>
    properties:
      hibernate:
        show-sql: true
        hbm2ddl:
          auto: none # <2>
        connection:
          url: ${datasources.default.url}
          username: ${datasources.default.username}
          password: ${datasources.default.password}
    reactive: true
---
datasources:
  default:
    driverClassName: com.mysql.cj.jdbc.Driver  # <3>
    dialect: MYSQL
flyway:
  datasources:
    default:
      enabled: true # <3>
```

And it seems to be working!

melix:
If you use latest test resources, you should even be able to simplify to:

```
jpa:
  default:
    db-type: mysql
    entity-scan:
      packages:
        - 'example.micronaut.domain' # <1>
    properties:
      hibernate:
        show-sql: true
        hbm2ddl:
          auto: update # <2>
    reactive: true
---
datasources:
  default:
    driverClassName: com.mysql.cj.jdbc.Driver  # <3>
    dialect: MYSQL
flyway:
  datasources:
    default:
      enabled: true # <3>
```

(note the `db-type` property)

timyates:
> (note the `db-type` property)

Do we then need DIALECT?

melix:
Fair point, no you shouldn't need the dialect, this passes tests:

```yaml
datasources:
  default:
    db-type: mysql
    schema-generate: CREATE_DROP
jpa:
  default:
    reactive: true
    entity-scan:
      packages: io.micronaut.testresources.hibernate.reactive.core
    properties:
      hibernate:
        hbm2ddl:
          auto: update
```

timyates:
Cool...  Right, we're in a bit of a feature/branch mess at the moment.

I will keep this PR JUST handling the non-test resources path.

And I will do the Test Resources changes to this as part of https://github.com/micronaut-projects/micronaut-starter/pull/1297

sdelamo (approved):

PR #1330, opened 2022-07-14
bug: Fix dependency exclusion for Gradle

timyates:
We used the wrong notation for exclusions in Gradle.

We had exclude(group: xxx, name: yyy) but it should be module instead of name

ie: exclude(group: xxx, module: yyy)

See https://docs.gradle.org/current/userguide/dependency_downgrade_and_exclude.html\#sec:excluding-transitive-deps

Bug was introduced here https://github.com/micronaut-projects/micronaut-starter/pull/1319 for 3.5.3, and was found when testing https://github.com/micronaut-projects/micronaut-guides/issues/1010

I've tagged this 3.5.4 (and based it off 3.5.x), but in all likelyhood it will be in 3.6.0


sdelamo (approved):

timyates:
@sdelamo ✅

PR #1340, opened 2022-07-27
Added micrometer-annotation feature

shingjo:
Feature for adding micrometer-annotation to starter.


starter-core/src/main/java/io/micronaut/starter/feature/micrometer/MicrometerAnnotations.java
```diff
@@ -0,0 +1,83 @@
...
+            featureContext.addFeature(management);
+        }
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.addDependency(Dependency.builder()
+                .groupId("io.micronaut.micrometer")
+                .artifactId("micronaut-micrometer-annotation")
+                .versionProperty("micronaut.micrometer.version")
```
sdelamo, line 70: is the property necessary?

shingjo: Yes, it is needed for Maven builds.

alvarosanchez: Then, don't you need to do it only if it's a Maven build?

shingjo: The full version of that section is:
```
generatorContext.addDependency(Dependency.builder()
                .groupId("io.micronaut.micrometer")
                .artifactId("micronaut-micrometer-annotation")
                .versionProperty("micronaut.micrometer.version")
                .annotationProcessor());
```

Without the `versionProperty` the Maven Rocker template fails to generate the pom correctly, because for annotations a version is required.

alvarosanchez: No, what I mean is that, don't you need to wrap that block in:

```java
if (!generatorContext.getBuildTool().isGradle()) {
  ...
}
```

shingjo: Alvaro, missed the point of your question. The version property seems to be only added for Maven and similar to other code in Starter.

alvarosanchez (approved):

PR #1366, opened 2022-08-15
feat: add MavenLocal repository

sdelamo:


timyates (approved):
I assume the usage of this is coming later?

PR #1368, opened 2022-08-15
extract getRepositories methods

sdelamo:
This allows to easily override `getRepositories` method to add extra repositories. e.g. `mavenLocal`


timyates (approved):

timyates:
re-running failed jobs as they should have passed

PR #1371, opened 2022-08-15
Generate test resources compatible configuration for R2DBC

### LINKED ISSUE #1353: Generate test resources compatible configuration for R2DBC
Current configuration is not compatible with test resources:

```yaml
r2dbc:
 datasources:
    default:
      schema-generate: CREATE_DROP
      dialect: H2
      url: r2dbc:h2:mem:///testdb;DB_CLOSE_ON_EXIT=FALSE
      username: sa
      password: ''

```

wetted:
fixes #1353


starter-core/src/test/groovy/io/micronaut/starter/feature/database/r2dbc/DataR2dbcSpec.groovy
```diff
@@ -177,16 +177,16 @@ class DataR2dbcSpec extends ApplicationContextSpec implements CommandOutputFixtu
         BuildTool.GRADLE_KOTLIN | "postgres"  | "POSTGRES"
         BuildTool.GRADLE        | "postgres"  | "POSTGRES"
         BuildTool.MAVEN         | "mysql"     | "MYSQL"
-        BuildTool.MAVEN         | "mysql"     | "MYSQL"
+        BuildTool.GRADLE_KOTLIN | "mysql"     | "MYSQL"
```
sdelamo: move this to https://github.com/micronaut-projects/micronaut-starter/pull/1374

starter-core/src/test/groovy/io/micronaut/starter/feature/testresources/DbTypeSpec.groovy
```diff
@@ -6,6 +6,14 @@ class DbTypeSpec extends Specification {
 
     void "toString() returns lowercase"() {
         expect:
-        "mariadb" == DbType.MARIADB.toString();
+        toString == dbtype.toString()
```
sdelamo: Move this to https://github.com/micronaut-projects/micronaut-starter/pull/1375

sdelamo (changes requested):
This still generates the following configuration: 

```
r2dbc:
  datasources:
    default:
      db-type: mysql
      url: r2dbc:mysql://localhost:3306/db
      username: root
      password: ''
```

I think we should generate something like: 
```
r2dbc:
  datasources:
    default:
      db-type: mysql
```      


@melix does Test Resources work with R2DBC ?

sdelamo:
We probably need to only add the properties only if TestResource feature is not present: 

https://github.com/micronaut-projects/micronaut-starter/blob/3.6.x/starter-core/src/main/java/io/micronaut/starter/feature/database/DatabaseDriverConfigurationFeature.java#L41-L45

Moreover, we will probably have to [extend from `EaseTestingFeature`](https://github.com/micronaut-projects/micronaut-starter/blob/3.6.x/starter-core/src/main/java/io/micronaut/starter/feature/testresources/EaseTestingFeature.java) to add the TestResource feature if the users select `r2dbc` feature. 

I don't think we need the validator. We should only add TestResource with a compatible database. The users don't apply `TestResource` feature themselves. 

And we need a test which verifies the username and password and url properties are not present in the database.

PR #1380, opened 2022-08-26
Update Camunda projects to versions built with Micronaut 3.6.x

tobiasschaefer:

Please merge this PR. It updates all features to versions built with Micronaut 3.6.x

FYI: @sdelamo

### DIFF
--- a/starter-core/src/main/resources/pom.xml
+++ b/starter-core/src/main/resources/pom.xml
@@ -80,17 +80,17 @@
         <dependency>
             <groupId>info.novatec</groupId>
             <artifactId>micronaut-camunda-bpm-feature</artifactId>
-            <version>2.8.0</version>
+            <version>2.9.0</version>
         </dependency>
         <dependency>
             <groupId>info.novatec</groupId>
             <artifactId>micronaut-camunda-external-client-feature</artifactId>
-            <version>2.7.1</version>
+            <version>2.8.0</version>
         </dependency>
         <dependency>
             <groupId>info.novatec</groupId>
             <artifactId>micronaut-zeebe-client-feature</artifactId>
-            <version>1.8.0</version>
+            <version>1.9.0</version>
         </dependency>
         <dependency>
             <groupId>org.jetbrains.kotlin</groupId>


alvarosanchez:
Wouldn't it be better for this feature to make the generated builds import `org.camunda.bpm:camunda-bom`? Then, we could get rid of those versions

tobiasschaefer:
Hi @alvarosanchez ,

thanks for your comment.

I cannot fully assess if it would be better to have the versions in the bom but it might even be better to not have them there (in our case).

If the versions were in the bom then this would apply: https://github.com/micronaut-projects/micronaut-core/wiki/Module-Upgrade-Strategy

However, as an open source project with limited resources we usually do the update to a new Micronaut major or minor version *after* it has been published (like in this PR we're updating to Micronaut 3.6 after its official release). If the limitations of the module upgrade strategy apply we wouldn't get this into the 3.6.x branch of micronaut-starter because we will always have at least a minor update (because Micronaut also has a minor update).

See also the discussion in:
https://github.com/micronaut-projects/micronaut-starter/pull/953

We've been doing updates like this PR now for almost two years which have mostly been merged by @sdelamo , see also https://github.com/micronaut-projects/micronaut-starter/pulls?q=is%3Apr+author%3Atobiasschaefer

For me it could stay like it has been. If we change it I need some support - but most importantly we need a solution where we can get our minor version update into the current release branch.

alvarosanchez:
What I propose would reduce the maintenance burden on your side. Let me discuss it with @sdelamo once he's back to work.

tobiasschaefer:
Hi @alvarosanchez ,

how would it reduce the maintainance burden? The version number would "only" be moved from the micronaut-starter project to micronout-core's bom, right?

Could we merge this PR so that the 3.6.x branch of micronaut-starter has the correct Camunda libraries if user create new applications? I assume the refactoring you're mentioning would not go into 3.6.x anyhow. So it makes sense to merge this now (preferably before the next 3.6.2 patch release)

sdelamo (commented):
I have merged them into the 3.7.x branch

tobiasschaefer:
Hi @sdelamo ,

thanks! Could you please also merge this to the 3.6.x branch so that the next patch release published on Micronaut Launch includes our releases which are built for Micronaut 3.6.x

PR #1387, opened 2022-09-05
feat: add versionCatalogue method

### LINKED ISSUE #1385: Use version catalogs in generated projects
Currently we use the "good old" dependency coordinates in generated templates. But instead, we could use the version catalog that we publish alongside the BOM.

For example, this is what the dependencies of a project with Micronaut Data and PostgreSQL would look like when using a catalog:

```gradle
dependencies {
    annotationProcessor(mn.micronaut.data.processor)
    annotationProcessor(mn.micronaut.http.validation)
    implementation(mn.micronaut.http.client)
    implementation(mn.micronaut.jackson.databind)
    implementation(mn.micronaut.data.jdbc)
    implementation(mn.micronaut.jdbc.hikari)
    implementation(mn.jakarta.annotation.api)
    runtimeOnly(mn.logback)
    runtimeOnly(mn.postgresql)
    implementation(mn.micronaut.validation)
}
```

Note that this requires updating our generated `settings.gradle` file:

```gradle
dependencyResolutionManagement {
   repositories {
       mavenCentral()
   }
   
   repositoriesMode = RepositoriesMode.PREFER_PROJECT
   
   versionCatalogs {
       create("mn") {
           from("io.micronaut:micronaut-bom:$micronautVersion")
       }
   }
}
```

Also for dependencies which are not in the published catalog, we'd have to add them in the `libs` catalog.

sdelamo:


starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleDependency.java
```diff
@@ -119,6 +122,14 @@ public String toSnippet() {
         return snippet;
     }
 
+    @NonNull
+    Optional<String> versionCatalogue() {
```
melix: I know you like the British version, but I think we should stick with the official terminology which is "version catalog" :)

sdelamo: resolved via https://github.com/micronaut-projects/micronaut-starter/pull/1387/commits/0f184501cebbf3992c826e1abf072570ea5e3a6b

melix (approved):

burtbeckwith:
@sdelamo these changes aren't backwards compatible. I don't think it's a big deal since starter libs aren't likely to be used in user apps, but an app that compiles with 3.6.x will not compile after this change. Consider adding overloads of `GradleBuildCreator.create()` and the `GradleDependency` constructor that pass `false` for `useVersionCatalogue`

alvarosanchez:
It is a big deal. This has broken micronaut-guides with 3.7.

sdelamo:
> @sdelamo these changes aren't backwards compatible. I don't think it's a big deal since starter libs aren't likely to be used in user apps, but an app that compiles with 3.6.x will not compile after this change. Consider adding overloads of `GradleBuildCreator.create()` and the `GradleDependency` constructor that pass `false` for `useVersionCatalogue`

I am not treating starter as other modules. I am introducing binary compatibility changes between minor versions. We can change that decision if we want. But this flexibility allows us to evolve Micronaut Guides faster.

PR #1391, opened 2022-09-06
Update Camunda projects to versions built with Micronaut 3.6.x

tobiasschaefer:

I recently created the PR #1380 which got merged into 3.7.x. Would it be possible to merge this PR as well into 3.6.x so that projects created with Micronaut Launch 3.6.x will have the correct Camunda projects releases.

Please merge this PR. It updates all features to versions built with Micronaut 3.6.x

FYI: @sdelamo

### DIFF
--- a/starter-core/src/main/resources/pom.xml
+++ b/starter-core/src/main/resources/pom.xml
@@ -80,17 +80,17 @@
         <dependency>
             <groupId>info.novatec</groupId>
             <artifactId>micronaut-camunda-bpm-feature</artifactId>
-            <version>2.8.0</version>
+            <version>2.9.0</version>
         </dependency>
         <dependency>
             <groupId>info.novatec</groupId>
             <artifactId>micronaut-camunda-external-client-feature</artifactId>
-            <version>2.7.1</version>
+            <version>2.8.0</version>
         </dependency>
         <dependency>
             <groupId>info.novatec</groupId>
             <artifactId>micronaut-zeebe-client-feature</artifactId>
-            <version>1.8.0</version>
+            <version>1.9.0</version>
         </dependency>
         <dependency>
             <groupId>org.jetbrains.kotlin</groupId>


sdelamo:
HI @tobiasschaefer we intentionally merge it into 3.7.x. We don't do minor upgrades in patch releases. 3.7.0 is planned for 15th september.

tobiasschaefer:
Hi @sdelamo ,

thanks for the Info. 15th September sounds good.

Please understand that with this approach the libraries we provide will always be at least one Micronaut minor release behind the version provided by Micronaut Launch. The reason being that we can only provide these once the Micronaut minor release has been published and according to your statement above you will not merge it into the current stable branch but in the upcoming release.

I would prefer to have it merged. You have been merging changes like this now for almost two years. Since it only affects new projects we won't break existing projects with minor Micronaut releases but we will make sure everything is on the same Micronaut version which has been tested with our CI. However, we don't test combinations for Micronaut Core 3.x and Micronaut Camunda 3.x-1.

Thanks for reading.

sdelamo:
> Please understand that with this approach the libraries we provide will always be at least one Micronaut minor release behind the version provided by Micronaut Launch. The reason being that we can only provide these once the Micronaut minor release has been published and according to your statement above you will not merge it into the current stable branch but in the upcoming release.
 
> I would prefer to have it merged. You have been merging changes like this now for almost two years. Since it only affects new projects we won't break existing projects with minor Micronaut releases but we will make sure everything is on the same Micronaut version which has been tested with our CI. However, we don't test combinations for Micronaut Core 3.x and Micronaut Camunda 3.x-1.


This is the approach we are also taking for internal modules. 

We are considering upgrading the Micronaut version to the next minor version, a change that requires the module to go into the next minor, and we don't include those in the BOM until the next minor version. 

For example: 

https://github.com/micronaut-projects/micronaut-rxjava2/pull/95

We are following the same strict approach for Micronaut Starter. We publish starter artifacts as well.

We may need to revisit this in the future. However, currently, the policy is that we don't include minor upgrades in patch versions of Micronaut starter. Sorry for the inconvenience.

PR #1396, opened 2022-09-12
ObjectStorageFeature should be OneOfFeature

sdelamo:
Maybe, we should make `ObjectStorageFeature` a `OneOfFeature`.


alvarosanchez:
I disagree. Object Storage is designed to be a multi-cloud API to allow users writing portable applications.

I admit that it is _unlikely_ that users are going to generate applications with more than one cloud, but there is also no reason for us to restrict them doing that.

PR #1400, opened 2022-09-13
refactor Gradle class to ease extensibility

sdelamo:


timyates (approved):

timyates:
Merged 3.7.x in to this to fix the conflict

Also, just reordered the methods so they follow the order they are called, my brain couldn't handle the jumping around 😉

PR #1418, opened 2022-09-15
Add a validator for the CRaC feature

timyates:
Checks we have not also included Graal, and that we are on Java 17+


sdelamo:
I am not sure about this `dockerBuildCrac` works without JDK 17

Maybe we could add a warning to the readme file if you are using something lower than JDK 17

timyates:
You're right, we don't require this

PR #1432, opened 2022-09-18
Maven test dependencies outside of rocker

sdelamo:
However, it does not use mockk version from the BOM for Maven Projects because that version does not work with Maven. 

Mockk has upgraded to use [Kotlin Multiplatform](https://github.com/mockk/mockk/issues/921) and does not work with Maven. 

I think we should add `io.mockk:mockk-jvm` to [micronaut-test](https://github.com/micronaut-projects/micronaut-test). Because of that, I think we should keep the same versions we used to have for now.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
@@ -147,14 +147,6 @@ MavenBuild mavenBuild
 @if (!features.contains("http-client")) {
 @dependency.template("io.micronaut", "micronaut-http-client", "test", null, false, null)
 }
-@if (features.testFramework().isJunit()) {
-    @if (features.contains("hamcrest")) {
-@dependency.template("org.hamcrest", "hamcrest", "test", null, false, null)
-    }
-    @if (features.contains("mockito")) {
-@dependency.template("org.mockito", "mockito-core", "test", null, false, null)
-    }
-}
 @if (features.contains("neo4j-bolt")) {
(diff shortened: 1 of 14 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
```diff
@@ -148,38 +148,12 @@ MavenBuild mavenBuild
...
 @if (features.testFramework().isJunit()) {
-@dependency.template("org.junit.jupiter", "junit-jupiter-api", "test", null, false, null)
-@dependency.template("org.junit.jupiter", "junit-jupiter-engine", "test", null, false, null)
-@dependency.template("io.micronaut.test", "micronaut-test-junit5", "test", null, false, null)
     @if (features.contains("hamcrest")) {
 @dependency.template("org.hamcrest", "hamcrest", "test", null, false, null)
     }
     @if (features.contains("mockito")) {
 @dependency.template("org.mockito", "mockito-core", "test", null, false, null)
     }
```
alvarosanchez: Why not moving these too? Then you could get rid of the whole `@if (features.testFramework().isJunit()) {}` block

alvarosanchez (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
alvarosanchez: Oh, never mind, I have just come across #1428

alvarosanchez (approved):

PR #1433, opened 2022-09-18
JUnit Params Feature

sdelamo:
Junit Params  provides easy parameterization of test methods in JUnit tests.


starter-core/src/main/resources/pom.xml
```diff
@@ -16,6 +16,11 @@
         </repository>
     </repositories>
     <dependencies>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>5.9.0</version>
```
alvarosanchez: This isn't needed. `micronaut-test-bom` imports `junit-bom`, which defines `junit-jupiter-params`

alvarosanchez (changes requested):

starter-core/src/main/resources/pom.xml
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-starter/pull/1433/commits/98fa17f537832b3f100bd87849a7afc3bb516112

alvarosanchez (approved):

PR #1466, opened 2022-10-06
feat: add command create-lambda

sdelamo:
This as an interactive prompt to create a AWS Lambda Function with Micronaut.

lambdagif


starter-cli/src/main/java/io/micronaut/starter/cli/command/CodingStyle.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.starter.cli.command;
+
+public enum CodingStyle {
+
+    CONTROLLERS("With @Controller classes, routes with methods annotated wtih @Get ..."),
```
alvarosanchez:
```suggestion
    CONTROLLERS("With @Controller classes, routes with methods annotated with @Get ..."),
```

starter-cli/src/main/java/io/micronaut/starter/cli/command/CodingStyle.java
```diff
@@ -0,0 +1,32 @@
...
+    HANDLER("Functional style with Handler class with input and output");
```
alvarosanchez:
```suggestion
    HANDLER("Functional style with a Handler class with input and output");
```

starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateLambdaBuilderCommand.java
```diff
@@ -0,0 +1,284 @@
...
+        out("Choose your trigger. (enter for " + defaultFeature.getTitle() + ")");
+        return getFeatureOption(
+                triggerFeatures(features),
+                Feature::getTitle,
+                defaultFeature,
+                reader);
+    }
+
+    protected Architecture getArchitecture(LineReader reader) {
+        out("Choose your Lambda Architecture. (enter for Java ARM)");
```
alvarosanchez, line 220: Is ARM the default because it's cheaper? Is it what most people actually use?

starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateLambdaBuilderCommand.java
```diff
@@ -0,0 +1,284 @@
...
+        return getEnumOption(
+                Architecture.class,
+                a -> a.toString() +  (a == Architecture.ARM ? " Better performance and 20% cheaper" : ""),
```
alvarosanchez: I'd remove the actual figure, since this can change in the future.
```suggestion
                a -> a.toString() +  (a == Architecture.ARM ? " Better performance and lower-priced" : ""),
```

starter-cli/src/main/java/io/micronaut/starter/cli/command/LambdaDeployment.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package io.micronaut.starter.cli.command;
+
+/**
+ * Lambda Deployments.
+ * @author Sergio del Amo
+ * @since 3.8.0
+ */
+public enum LambdaDeployment {
+    FAT_JAR("FAT JAR running in the managed Java Runtime."),
```
alvarosanchez:
```suggestion
    FAT_JAR("Fat JAR running in the managed Java runtime"),
```

starter-cli/src/main/java/io/micronaut/starter/cli/command/LambdaDeployment.java
```diff
@@ -0,0 +1,36 @@
...
+    NATIVE_EXECUTABLE("Native executable built with GraalVM Native Image running in Custom Runtime. Faster Cold Start");
```
alvarosanchez:
```suggestion
    NATIVE_EXECUTABLE("Native executable built with GraalVM Native Image running in a custom runtime. Faster cold start");
```

starter-core/src/main/java/io/micronaut/starter/feature/architecture/Arm.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package io.micronaut.starter.feature.architecture;
+
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.starter.feature.aws.Cdk;
+import io.micronaut.starter.feature.function.awslambda.AwsLambda;
+import jakarta.inject.Singleton;
+
+@Singleton
+public class Arm implements CpuArchitecture {
```
alvarosanchez, line 24: Why they need to be features? Can't we just use the `Architecture` enum? Pretty much as we do with `JdkVersion` for example.

starter-core/src/main/java/io/micronaut/starter/feature/aws/template/cdkappstack.rocker.raw
```diff
@@ -140,6 +142,11 @@ public class AppStack extends Stack {
                 .memorySize(512)
                 .tracing(Tracing.ACTIVE)
                 .logRetention(RetentionDays.ONE_WEEK)
+@if (architecture == Architecture.ARM) {
+                .architecture(Architecture.ARM_64)
+} else {
```
alvarosanchez: See https://github.com/micronaut-projects/micronaut-starter/blob/3.7.x/CONTRIBUTING.md#avoid-empty-else and avoid this empty else.

starter-core/src/main/java/io/micronaut/starter/feature/graalvm/GraalVMFeatureValidator.java
```diff
@@ -39,7 +39,7 @@ public void validatePreProcessing(Options options, ApplicationType applicationTy
     @Override
     public void validatePostProcessing(Options options, ApplicationType applicationType, Set<Feature> features) {
         if (features.stream().anyMatch(f -> f instanceof GraalVM || f instanceof GraalVMDockerRegistryWorkflow)) {
-            if (options.getLanguage() == Language.GROOVY) {
+            if (!supports(options.getLanguage())) {
                 throw new IllegalArgumentException("GraalVM is not supported in Groovy applications");
             }
```
alvarosanchez: Refactor to avoid having Groovy hardcoded in the condition is great. But then, the exception message is still hardcoded. This change is not worth it IMO

alvarosanchez (changes requested):

starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateLambdaBuilderCommand.java
sdelamo: Yes, it is what they recommend. It is cheaper and faster. You should only use x86 if you are in AWS region where they don't have ARM yet.

starter-core/src/main/java/io/micronaut/starter/feature/graalvm/GraalVMFeatureValidator.java
sdelamo: > Refactor to avoid having Groovy hardcoded in the condition is great. But then, the exception message is still hardcoded

Good catch. I missed the harcoded string. I have removed it here: 

https://github.com/micronaut-projects/micronaut-starter/pull/1466/commits/5c24c4d0ea604521a532e8e7f153be948fe654c4

starter-core/src/main/java/io/micronaut/starter/feature/aws/template/cdkappstack.rocker.raw
sdelamo: Resolved via https://github.com/micronaut-projects/micronaut-starter/pull/1466/commits/49da2f8f0ea614d25759c48846db10cb46d2a88a

starter-core/src/main/java/io/micronaut/starter/feature/architecture/Arm.java
sdelamo: > Why they need to be features? Can't we use the `Architecture` enum? Pretty much as we do with `JdkVersion`, for example.

We could do it without features and add it to Options as you suggest.
https://github.com/micronaut-projects/micronaut-starter/blob/3.7.x/starter-core/src/main/java/io/micronaut/starter/options/Options.java

However, currently, this selection only influences a cloud vendor service. I think it is more flexible and unobtrusive as a feature. 

I would suggest leaving it as a feature for now but no strong opinion about this. I see your point.

sdelamo: Another advantage of using it as a feature is that we don't need to modify the front end to support it immediately.

alvarosanchez: Ok, fair enough, but then please remove the enum. `CpuArchitecture` is already a one-of feature. Note that currently for 2 options we have:
- `enum Architecture`.
- `class CpuArchitecture`.
- `class Arm`.
- `class X86`.

IMHO it is unnecessary complex.

alvarosanchez (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/architecture/Arm.java
sdelamo: I have removed the enum. Removing it has forced me to do several `instanceof` and add `getListOption` method.  I think the code was simpler with the enum.  🤷🏻‍♂️

https://github.com/micronaut-projects/micronaut-starter/pull/1466/commits/5c24c4d0ea604521a532e8e7f153be948fe654c4

alvarosanchez (approved):
There are build failures, however

sdelamo:
```
sdk use java 22.2.r11-grl
./gradlew micronaut-cli:build       
native-image --no-fallback -cp starter-cli/build/libs/micronaut-cli-*-all.jar
```

fails with: 
```
Fatal error: java.lang.IllegalAccessError: class io.micronaut.buffer.netty.NettyFeature (in unnamed module @0x33eb0d4) cannot access class com.oracle.svm.core.jdk.SystemPropertiesSupport (in module org.graalvm.nativeimage.builder) because module org.graalvm.nativeimage.builder does not export com.oracle.svm.core.jdk to unnamed module @0x33eb0d4
	at io.micronaut.buffer.netty.NettyFeature.beforeAnalysis(NettyFeature.java:66)
```

I don't think it is related.

alvarosanchez:
That failure rings a bell. @melix didn't we fix that in some version of the Gradle plugin?

melix:
mmm, yes, I'm pretty sure this error is fixed in latest Gradle plugin releases. Which version of the plugin this project is using? Got a build scan? However that's not the same error we have in CI.

graemerocher:
Should be called `create-aws-lambda` imo

alvarosanchez:
@melix see #1469

sdelamo:
> Should be called `create-aws-lambda` imo

I have renamed it https://github.com/micronaut-projects/micronaut-starter/pull/1466/commits/988f56637e6c7c2e828693102b39c7c0f31373c6

timyates:
Pushed a fix for a maven test that was failing as the base image is now `ol7-java11-22.3.0` not `ol7-java11-22.2.0`

https://github.com/micronaut-projects/micronaut-starter/pull/1466/commits/6216470c42683cb8d44ea6a045699e2db23c09b1

PR #1484, opened 2022-10-18
feat: don't enable jansi for aws-lambda feature

sdelamo:


timyates (approved):
I know what this is, as I was in London when you explained it 😎

alvarosanchez (approved):

PR #1505, opened 2022-10-27
build: GraalVM to 22.3

sdelamo:


alvarosanchez (approved):
All the workflows are still using 21.3.0. Should we change that to `latest`?

PR #1506, opened 2022-10-28
Fix TestResources with Hibernate Reactive

timyates:
If we are using hibernate-reactive and Test Resources, we need to specify the correct module test resources to use. We also need to provide the JDBC driver for the chosen database as a `testResourcesService` dependency.

In Maven, this is done by adding both to the `testResourcesDependencies` block of the `micronaut-maven-plugin` block

Whilst doing this, I also discovered we were using the wrong `db-type` for postgres (we had `postgresql`, but it should be `postgres`)

We have an issue with using the vert.x drives with postgres, as can be seen in the feature tests (I commented postgres out), I will raise a separate issue for this.

I have tagged this as being for 3.7.3 but it's a bit of a big change, and I wonder if 3.7.4 after some thorough testing for edge cases would be appropriate


timyates:
Raised https://github.com/micronaut-projects/micronaut-sql/issues/742 for the postgres issue

test-features/src/test/groovy/io/micronaut/starter/core/test/feature/database/HibernateReactiveJpaSpec.groovy
```diff
@@ -0,0 +1,77 @@
...
+        String output = executeMaven("compile test")
+
+        then:
+        output?.contains("BUILD SUCCESS")
+
+        where:
+        db << [
+                MySQL.NAME,
+                MariaDB.NAME,
+//                PostgreSQL.NAME, // TODO: Reinstate Postgres test once the vert.x driver missing dependency is sorted out
```
melix: So you haven't found the magic trick... :thinking:

melix (approved):

test-features/src/test/groovy/io/micronaut/starter/core/test/feature/database/HibernateReactiveJpaSpec.groovy
timyates: I'm wondering whether we just modify starter to pull in the `com.ongres.scram:client` dependency if postgres and reactive hibernate are selected

timyates: Added the dependency as part of this, and re-enabled Postgres in the feature test

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
```diff
@@ -167,23 +173,31 @@ MavenBuild mavenBuild
...
-    .map(dbType -> features.isFeaturePresent(R2dbc.class) ? dbType.getR2dbcTestResourcesModuleName() : dbType.getJdbcTestResourcesModuleName())
+    .map(dbType -> features.isFeaturePresent(R2dbc.class) ? dbType.getR2dbcTestResourcesModuleName() : features.isFeaturePresent(HibernateReactiveFeature.class) ? dbType.getHibernateReactiveTestResourcesModuleName() : dbType.getJdbcTestResourcesModuleName())
     .orElse(null)) {
             <classpathInference>false</classpathInference>
             <testResourcesDependencies>
               <dependency>
                 <groupId>io.micronaut.testresources</groupId>
                 <artifactId>micronaut-test-resources-@resourceName</artifactId>
               </dependency>
+@if (features.isFeaturePresent(HibernateReactiveFeature.class) && features.isFeaturePresent(DatabaseDriverFeature.class) && !features.isFeaturePresent(MigrationFeature.class)) {
```
sdelamo, line 193: We should not add extra logic to pom.rocker if possible

sdelamo (approved):

PR #1515, opened 2022-11-03
Migrate to MN4

wetted:
This seems to fix some of the errors building starter, but there are still a couple annotation processing and  build errors I haven't figured out yet.

One is coming out the annotation processor for `:starter-analytics-postgres:compileJava`. 

```
:starter-analytics-postgres:compileJava FAILED	
error: Unable to implement Repository method: FeatureRepository.save(Object entity). No possible implementations found.	
Note: Creating bean classes for 6 type elements
```

The other is a groovy codehaus/apache collision. I added a dependency substitution to the buildSrc 'base' conventions plugin, but that isn't helping for this particular error.

```
> Task :test-utils:compileGroovy FAILED
> startup failed:
> Could not instantiate global transform class org.spockframework.compiler.SpockTransform specified at jar:file:/Users/wetted/.gradle/caches/modules-2/files-2.1/org.spockframework/spock-core/2.3-groovy-4.0/8861b2590bb8e4709b052fb4ed6da3de98e734d9/spock-core-2.3-groovy-4.0.jar!/META-INF/services/org.codehaus.groovy.transform.ASTTransformation  because of exception org.spockframework.util.IncompatibleGroovyVersionException: The Spock compiler plugin cannot execute because Spock 2.3.0-groovy-4.0 is not compatible with Groovy 3.0.10. For more information (including enforce mode), see https://docs.spockframework.org (section 'Known Issues').
> Spock artifact: file:/Users/wetted/.gradle/caches/modules-2/files-2.1/org.spockframework/spock-core/2.3-groovy-4.0/8861b2590bb8e4709b052fb4ed6da3de98e734d9/spock-core-2.3-groovy-4.0.jar
> Groovy artifact: file:/Users/wetted/.gradle/wrapper/dists/gradle-7.5.1-bin/7jzzequgds1hbszbhq3npc5ng/gradle-7.5.1/lib/groovy-3.0.10.jar
```


starter-core/build.gradle
```diff
@@ -18,6 +18,7 @@ dependencies {
     api "io.micronaut:micronaut-inject"
     api "io.micronaut:micronaut-http"
     api "io.micronaut:micronaut-http-client"
+    api "io.micronaut:micronaut-jackson-databind"
```
graemerocher: can we move to Micronaut Serialization?

wetted:
This also needs more work resolving the serialization stuff too. I'm on it.

wetted:
The project builds now, but the tests don't pass, since they need to point to module snapshots for 4.0.x that resolve the ConversionService changes (and maybe other things too).

wetted:
Still having issues with `org.codehaus.groovy` groovy 3 leaking into build and causing groovy compile failures. Any idea how to solve this @melix? When I look at the build scan I see no transitive module dependencies on the old groovy, but the build plugin has such a dependency. I tried some hacks and removed them when they didn't work. I upgraded to the project build to gradle 8.0.1 to see if that would help somehow.

PR #1526, opened 2022-11-10
Enable eager initialization of singletons with CRaC

### LINKED ISSUE #1513: when selecting crac feature we should set eager intiailization of singletons
https://docs.micronaut.io/latest/guide/#eagerInit

it is beneficial to eagerly initialise every singleton prior to checkpoint

timyates:
This PR sets the eager initialization of singletons option if the CRaC feature is selected

https://docs.micronaut.io/latest/guide/#eagerInit

To allow this, we had to switch from `Micronaut.run` to `Micronaut.build` in the Application main class.

Fixes #1513

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
@@ -1,5 +1,6 @@
 @import io.micronaut.starter.application.Project
 @import io.micronaut.starter.feature.Features
+@import io.micronaut.starter.feature.crac.Crac
 
 @args (Project project, Features features, String defaultEnvironment)
 
@@ -34,11 +35,19 @@ class Application {
     static class DefaultEnvironmentConfigurer implements ApplicationContextConfigurer {
         @@Override
         public void configure(@@NonNull ApplicationContextBuilder builder) {
-            builder.defaultEnvironments("@(defaultEnvironment)");
+            builder.defaultEnvironments("@(defaultEnvironment)")
+@if (features.contains(Crac.NAME)) {
+            builder.eagerInitSingletons(true)
+}
         }
     }
 }
     static void main(String[] args) {
-        Micronaut.run(Application, args)
+        Micronaut.build(args)
+            .mainClass(Application.class)
+@if (features.contains(Crac.NAME)) {
+            .eagerInitSingletons(true)
+}
+            .start()
     }
 }
--- a/starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
@@ -1,8 +1,7 @@
 @import io.micronaut.starter.application.Project
 @import io.micronaut.starter.feature.Features
-@import io.micronaut.starter.feature.crac.Crac
 
-@args (Project project, Features features, String defaultEnvironment)
+@args (Project project, Features features, String defaultEnvironment, boolean eagerInitSingleton)
 
 package @project.getPackageName()
 
@@ -36,7 +35,7 @@ class Application {
         @@Override
         public void configure(@@NonNull ApplicationContextBuilder builder) {
             builder.defaultEnvironments("@(defaultEnvironment)")
-@if (features.contains(Crac.NAME)) {
+@if (eagerInitSingleton) {
             builder.eagerInitSingletons(true)
 }
         }
@@ -45,7 +44,7 @@ class Application {
     static void main(String[] args) {
         Micronaut.build(args)
             .mainClass(Application.class)
-@if (features.contains(Crac.NAME)) {
+@if (eagerInitSingleton) {
             .eagerInitSingletons(true)
 }
             .start()
--- a/starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
@@ -1,11 +1,12 @@
 @import io.micronaut.starter.application.Project
 @import io.micronaut.starter.feature.Features
+@import io.micronaut.starter.feature.lang.ApplicationRenderingContext
 
-@args (Project project, Features features, String defaultEnvironment, boolean eagerInitSingleton)
+@args (Project project, Features features, ApplicationRenderingContext context)
 
 package @project.getPackageName()
 
-@if (defaultEnvironment != null) {
+@if (context.isRequired()) {
 import io.micronaut.core.annotation.NonNull
 import io.micronaut.context.ApplicationContextBuilder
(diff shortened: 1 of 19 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
```diff
@@ -34,11 +35,19 @@ class Application {
     static class DefaultEnvironmentConfigurer implements ApplicationContextConfigurer {
         @@Override
         public void configure(@@NonNull ApplicationContextBuilder builder) {
-            builder.defaultEnvironments("@(defaultEnvironment)");
+            builder.defaultEnvironments("@(defaultEnvironment)")
+@if (features.contains(Crac.NAME)) {
```
sdelamo: I would prefer if we pass a boolean with `eagerInitSingleton` and remove features logic from rocker templates

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
timyates: I created an interface `RequireEagerSingletonInitializationFeature` which extends `Feature` and switched `Crac` to implement that instead

We then check for any `RequireEagerSingletonInitializationFeature` features, and pass a boolean to the template

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
```diff
@@ -34,11 +34,19 @@ class Application {
     static class DefaultEnvironmentConfigurer implements ApplicationContextConfigurer {
```
sdelamo: I think we should rename static  this to be more agnostic. Please, note you could have one feature which requires eager initialization of singletons but does not specify a default environment.

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
```diff
@@ -34,11 +34,19 @@ class Application {
...
         @@Override
         public void configure(@@NonNull ApplicationContextBuilder builder) {
-            builder.defaultEnvironments("@(defaultEnvironment)");
+            builder.defaultEnvironments("@(defaultEnvironment)")
+@if (eagerInitSingleton) {
```
sdelamo: I think it is best to remove this. We don't have any AOT feature which requires `eagerInitialization` to be specified via an `ApplicationContextConfigurer`.

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
sdelamo: or just remove eager initialization from the ApplicationContextConfigurer as specified below

starter-core/src/main/java/io/micronaut/starter/feature/lang/java/application.rocker.raw
```diff
@@ -86,10 +86,18 @@ public class Application {
         @@Override
         public void configure(@@NonNull ApplicationContextBuilder builder) {
             builder.defaultEnvironments("@(defaultEnvironment)");
+@if (eagerInitSingleton) {
```
sdelamo: remove it from here?

starter-core/src/main/java/io/micronaut/starter/feature/lang/kotlin/application.rocker.raw
```diff
@@ -90,13 +90,20 @@ object Dekorate {
 
 @@ContextConfigurer
 class DefaultEnvironmentConfigurer: ApplicationContextConfigurer {
-	override fun configure(builder: ApplicationContextBuilder) {
-		builder.defaultEnvironments("@(defaultEnvironment)");
+    override fun configure(builder: ApplicationContextBuilder) {
+        builder.defaultEnvironments("@(defaultEnvironment)");
+@if (eagerInitSingleton) {
```
sdelamo: remove it from here?

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
```diff
@@ -34,11 +34,19 @@ class Application {
...
+            builder.eagerInitSingletons(true)
+}
         }
     }
 }
     static void main(String[] args) {
-        Micronaut.run(Application, args)
```
sdelamo, line 42: change to `build().start()` instead of `run`only if `eagerInitSingleton` is `true`.

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
sdelamo: ping @melix

timyates:
Added @melix as a reviewer as I believe the `ContextConfigurer` is for AOT?

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/application.rocker.raw
timyates: Switched to just use run all the time, and add a context configurer if there's a default environment OR a eager singleton requirement

melix (approved):
Use of the context configurer looks good to me. I am simply wondering about the class name, `CustomApplicationContextConfigurer` which is a bit long. Since we don't really care about the name, we could simply call it `Configurer`, for example.

timyates:
@sdelamo do you agree on `Configurer` for the name?  If so, I'll change it 👍

sdelamo:
> @sdelamo do you agree on `Configurer` for the name? If so, I'll change it 👍

Yes, that it is fine.

sdelamo (changes requested):
@timyates can you add a functional tests so we ensure that Java, Groovy, Kotlin code compiles fine.

sdelamo:
> @timyates can you add a functional tests so we ensure that Java, Groovy, Kotlin code compiles fine.

or is it already covered by CRaC feature tests?

timyates:
@sdelamo Yeah, that's covered by the feature test in https://github.com/micronaut-projects/micronaut-starter/pull/1526/files#diff-1c9947fca60ad7ce66e177023fd7abcf936bff6ccdb3914b16a4aa7b4cb4a965

timyates:
> > @sdelamo do you agree on `Configurer` for the name? If so, I'll change it 👍
> 
> Yes, that it is fine.

Done in https://github.com/micronaut-projects/micronaut-starter/pull/1526/commits/1c2ccde0114feef2fe658e45441040b41b7bee95

sdelamo (approved):

timyates:
Fixing tests...

PR #1528, opened 2022-11-10
Switch to enable classpathInference for test resources

### LINKED ISSUE #1522: Mysql Maven project doesn't work out of the box with TestResources
If you create a maven app with the mysql feature, the mysql driver doesn't end up on the classpath of the test-resources

```
Caused by: java.lang.ClassNotFoundException: com.mysql.jdbc.Driver
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:315)
        at org.testcontainers.containers.JdbcDatabaseContainer.getJdbcDriverInstance(JdbcDatabaseContainer.java:200)
        ... 83 common frames omitted
```

The workaround is to change the maven plugin config from

```
          <configuration>
            <classpathInference>false</classpathInference>
```

To

```
          <configuration>
            <classpathInference>true</classpathInference>
```

@alvarosanchez should we be setting this to `true` by default for everything using test-resources?

Micronaut 3.7.3

timyates:
It was noticed in #1522 that a plain mysql app didn't work as the TR service didn't have the driver class available.

We decided to enable classpathInference as that works, and that is how the Gradle plugin works also

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/testresources/EaseTestingFeature.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/testresources/EaseTestingFeature.java
@@ -36,7 +36,7 @@ public abstract class EaseTestingFeature implements Feature  {
 
     @Override
     public void processSelectedFeatures(FeatureContext featureContext) {
-        if (!featureContext.isPresent(TestResources.class) && testResources != null) {
+        if (!featureContext.isPresent(TestContainers.class) && testResources != null) {
             featureContext.addFeature(testResources);
         }
     }
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw
@@ -184,7 +184,7 @@ MavenBuild mavenBuild
     .flatMap(DatabaseDriverFeature::getDbType)
     .map(dbType -> features.isFeaturePresent(R2dbc.class) ? dbType.getR2dbcTestResourcesModuleName() : features.isFeaturePresent(HibernateReactiveFeature.class) ? dbType.getHibernateReactiveTestResourcesModuleName() : dbType.getJdbcTestResourcesModuleName())
     .orElse(null)) {
-            <classpathInference>false</classpathInference>
+            <classpathInference>true</classpathInference>
             <testResourcesDependencies>
               <dependency>
                 <groupId>io.micronaut.testresources</groupId>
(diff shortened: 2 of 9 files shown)


alvarosanchez (changes requested):
`classpathInference` defaults to true, so simply remove the tag.

Also, `testResourcesDependencies` is only to be used when NOT using classpath inference.

timyates:
> Also, `testResourcesDependencies` is only to be used when NOT using classpath inference.

Is that true?  I thought we had to add the test-resources module to `testResourcesDependencies`, and if we're using reactive sql we need to add the jdbc-driver too?

alvarosanchez:
> Is that true? I thought we had to add the test-resources module to `testResourcesDependencies`, and if we're using reactive sql we need to add the jdbc-driver too?

We have probably been doing this wrong, as @melix [clarified](https://github.com/micronaut-projects/micronaut-starter/issues/1522#issuecomment-1310316056). `io.micronaut.testresources:micronaut-test-resources-jdbc-mysql` will definitely be inferred automatically.

timyates:
> We have probably been doing this wrong, as @melix https://github.com/micronaut-projects/micronaut-starter/issues/1522#issuecomment-1310316056. io.micronaut.testresources:micronaut-test-resources-jdbc-mysql will definitely be inferred automatically.

But we still need to add the mysql driver if for example they are using the vertx reactive one in their code?

alvarosanchez:
> But we still need to add the mysql driver if for example they are using the vertx reactive one in their code?

I guess so. But in your original comment you mentioned a plain mysql app. The `mysql` feature will add the driver to the classpath

timyates:
Sorry, I was talking about https://github.com/micronaut-projects/micronaut-starter/pull/1528#issuecomment-1311893057

I'll take a look

timyates:
@melix do we have a list of rules as to when the `micronaut-test-resources-@resourceName` module is required?

I'm trying to extract this from Rocker and into java

https://github.com/micronaut-projects/micronaut-starter/blob/dd21765c5d1441e1320af425ff7a36b3d2b52358/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw#L181-L201

timyates:
Found another issue...

If `TestContainers` is selected, it's supposed to disable `TestResources` 🤔

starter-core/src/main/java/io/micronaut/starter/feature/testresources/EaseTestingFeature.java
```diff
@@ -36,7 +36,7 @@ protected EaseTestingFeature(TestContainers testContainers,
 
     @Override
     public void processSelectedFeatures(FeatureContext featureContext) {
-        if (!featureContext.isPresent(TestResources.class) && testResources != null) {
+        if (!featureContext.isPresent(TestContainers.class) && testResources != null) {
```
alvarosanchez, line 39: Are you sure this is correct? I think the original intent was to check the presence of `TestResources` before adding it. It's an idiom repeated all over starter.

timyates: The requirement is:

If they have selected `TestContainers` then don't apply `TestResources`

If they have selected nothing, apply `TestResources`

Previously we applied `TestResources` all the time, whether they had added `TestContainers` or not 🤔 

Do you have a link to the idiom?

alvarosanchez (approved):

sdelamo (approved):

PR #1537, opened 2022-11-16
Don't set `datasources.default.schema-generate: CREATE_DROP` when adding liquibase or flyway feature

### LINKED ISSUE #1524: Don't set datasources.default. schema-generate: CREATE_DROP when adding flyway
Currently for example for `data-jdbc` and `flyway` we generate. 

```
datasources:
  default:
    url: jdbc:h2:mem:devDb;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
    username: sa
    password: ''
    driverClassName: org.h2.Driver
    schema-generate: CREATE_DROP
    dialect: H2
```

I think we should generate: 

```
datasources:
  default:
    url: jdbc:h2:mem:devDb;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
    username: sa
    password: ''
    driverClassName: org.h2.Driver
    dialect: H2
flyway:
  datasources: 
    default: 
      enabled: true 
```

Thus, removing `datasources.default. schema-generate: CREATE_DROP`

### LINKED ISSUE #1525: Don't set datasources.default. schema-generate: CREATE_DROP when adding liquibase
Currently for example for `data-jdbc` and `flyway` we generate. 

```
datasources:
  default:
    url: jdbc:h2:mem:devDb;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
    username: sa
    password: ''
    driverClassName: org.h2.Driver
    schema-generate: CREATE_DROP
    dialect: H2
```

I think we should generate: 

```
datasources:
  default:
    url: jdbc:h2:mem:devDb;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
    username: sa
    password: ''
    driverClassName: org.h2.Driver
    dialect: H2
liquibase:
    datasources: 
        default: 
            change-log: 'classpath:db/liquibase-changelog.xml'
```

Thus, removing `datasources.default. schema-generate: CREATE_DROP`

we should generate a `src/main/resources/db/liquibase-changelog.xml` file with content: 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<databaseChangeLog
  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
  <include file="changelog/01-schema.xml" relativeToChangelogFile="true"/>
</databaseChangeLog>
```

and a `src/main/resources/db/01-schema.xml`

```
<?xml version="1.0" encoding="UTF-8"?>

<databaseChangeLog
  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
  <changeSet id="01" author="username">
  </changeSet>
</databaseChangeLog>
```

wetted:
closes #1524
closes #1525

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/database/DataJdbc.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/database/DataJdbc.java
@@ -19,10 +19,7 @@ import io.micronaut.starter.application.generator.GeneratorContext;
 import io.micronaut.starter.build.dependencies.Dependency;
 import io.micronaut.starter.feature.FeatureContext;
 import io.micronaut.starter.feature.database.jdbc.JdbcFeature;
-import io.micronaut.starter.feature.migration.MigrationFeature;
 import jakarta.inject.Singleton;
-
-import java.util.Map;
 
 @Singleton
 public class DataJdbc implements DataFeature {
@@ -68,11 +65,7 @@ public class DataJdbc implements DataFeature {
                 .compile());
 
         DatabaseDriverFeature dbFeature = generatorContext.getRequiredFeature(DatabaseDriverFeature.class);
-        Map<String, Object> datasourceConfig = getDatasourceConfig(dbFeature);
-        if (generatorContext.isFeaturePresent(MigrationFeature.class)) {
-            datasourceConfig.remove(SCHEMA_GENERATE_KEY);
-        }
-        generatorContext.getConfiguration().addNested(datasourceConfig);
+        generatorContext.getConfiguration().addNested(getDatasourceConfig(generatorContext, dbFeature));
     }
 
     @Override
--- a/starter-core/src/main/java/io/micronaut/starter/feature/database/DataJdbc.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/database/DataJdbc.java
@@ -19,8 +19,10 @@ import io.micronaut.starter.application.generator.GeneratorContext;
 import io.micronaut.starter.build.dependencies.Dependency;
 import io.micronaut.starter.feature.FeatureContext;
 import io.micronaut.starter.feature.database.jdbc.JdbcFeature;
-
+import io.micronaut.starter.feature.migration.MigrationFeature;
 import jakarta.inject.Singleton;
+
+import java.util.Map;
 
 @Singleton
 public class DataJdbc implements DataFeature {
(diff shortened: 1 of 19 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/database/DataJdbc.java
```diff
@@ -64,8 +66,13 @@ public void apply(GeneratorContext generatorContext) {
                 .groupId("io.micronaut.data")
                 .artifactId("micronaut-data-jdbc")
                 .compile());
+
         DatabaseDriverFeature dbFeature = generatorContext.getRequiredFeature(DatabaseDriverFeature.class);
-        generatorContext.getConfiguration().addNested(getDatasourceConfig(dbFeature));
+        Map<String, Object> datasourceConfig = getDatasourceConfig(dbFeature);
```
timyates: Instead of adding it in `getDataSourceConfig` and then removing it, do you think it would be better to pass the generator context to `getDataSourceConfig` and don't add the config if a Migration feature is present there?

wetted: I have mixed feelings about this. I avoid changing a API unless there's a compelling reason to do so. If I change as you suggested, then knowledge of the `Migration` feature is pushed up to the `DataFeature` interface, but that's relevant to just a few implementations, two of which already override the method. `DataJpa` overrides the `getDataSourceConfig` in its implementation, to never add that conditional config. `DataR2dbc` overrides it as well since the config is different for it; thus, the `Migration` logic has to be tested in the implementation class anyway.

I have a pushed a commit that makes your suggested change. See what you think. Thoughts @sdelamo?

starter-core/src/main/java/io/micronaut/starter/feature/migration/Liquibase.java
```diff
@@ -50,9 +53,15 @@ public String getMicronautDocumentation() {
 
     @Override
     public void apply(GeneratorContext generatorContext) {
+        generatorContext.addTemplate("liquibaseChangelog", new RockerTemplate("src/main/resources/db/liquibase-changelog.xml",
+                        liquibaseChangelog.template()));
+        generatorContext.addTemplate("liquibaseSchema", new RockerTemplate("src/main/resources/db/01-schema.xml",
```
sdelamo: this is wrong. it should be within the `src/main/resources/db/changelog` folder

sdelamo (changes requested):

sdelamo (approved):

PR #1552, opened 2022-11-26
Gradle 7.6.1

### LINKED ISSUE #1551: Upgrade to Gradle 7.6
We have to do it in two places:  https://github.com/micronaut-projects/micronaut-starter/pull/1373/files

timyates:


sdelamo (changes requested):
There should be 8 files changed. 4 in root project itself, 4 in `starter-core/src/main/resources/gradle`

timyates:
Also `HibernateReactiveJpaSpec` seems to be hanging... Maybe related

sdelamo (approved):

timyates:
Right...it's the Oracle test that's hanging...

Investigating on an x86 machine where the image works...

timyates:
So both Maven and Gradle hang when doing the oracle feature test for Hibernate Reactive JPA... 

I am currently stumped as to why...

PR #1580, opened 2023-01-11
Remove aws-java-sdk-core as a test dependency

timyates:
When we add aws-java-sdk-core as a test dependency, Maven removes it from the compile time transitive dependencies.

This causes the jms-aws-sqs maven guide to fail as it cannot find com.amazonaws.regions.Regions at compile time.

As the dependency comes in transitively, we can remove it from the locastack feature, as it should be coming in transitively from whichever micronaut aws library the user is using


timyates:
Converted to Draft whilst @alvarosanchez investigates

alvarosanchez (approved):
LGTM

sdelamo (approved):

PR #1584, opened 2023-01-12
Feature for data-azure-cosmos

### LINKED ISSUE #1573: Add feature for Micronaut Data Cosmos
see: https://micronaut-projects.github.io/micronaut-data/latest/guide/#azureCosmos

radovanradic:
Close #1573


radovanradic:
#1573

starter-core/src/main/java/io/micronaut/starter/feature/database/DataAzureCosmosFeature.java
```diff
@@ -0,0 +1,120 @@
...
+public class DataAzureCosmosFeature implements DataFeature {
+
+    private static final String MICRONAUT_DATA_GROUP = "io.micronaut.data";
+    private static final String MICRONAUT_DATA_VERSION = "micronaut.data.version";
+    private static final String MICRONAUT_DATA_PROCESSOR_ARTIFACT = "micronaut-data-processor";
+    private static final String MICRONAUT_DATA_DOCUMENT_PROCESSOR_ARTIFACT = "micronaut-data-document-processor";
+    private static final String MICRONAUT_DATA_AZURE_COSMOS_DATA_ARTIFACT = "micronaut-data-azure-cosmos";
+
+    // These are just placeholder values for user to replace
+    private static final String DEFAULT_ENDPOINT = "https://some-host/some-db";
```
sdelamo: I don't know about generating configuration with placeholders. We typically don't do that for other applications, and when we do, we use valid placeholders. If we keep this when a user generates an app. ./gradlew test will fail with:

```
io.micronaut.context.exceptions.BeanInstantiationException: Bean definition [io.micronaut.data.cosmos.common.CosmosDatabaseInitializer] could not be loaded: Error instantiating bean of type  [io.micronaut.data.cosmos.common.CosmosDatabaseInitializer]

Message: Client initialization failed. Check if the endpoint is reachable and if your auth token is valid. More info: https://aka.ms/cosmosdb-tsg-service-unavailable-java
Path Taken: new CosmosDatabaseInitializer() --> CosmosDatabaseInitializer.initialize([CosmosClient cosmosClient],RuntimeEntityRegistry runtimeEntityRegistry,CosmosDiagnosticsProcessor cosmosDiagnosticsProcessor,CosmosDatabaseConfiguration configuration)
	at app//io.micronaut.context.DefaultBeanContext.initializeContext(DefaultBeanContext.java:1921)
	at app//io.micronaut.context.DefaultApplicationContext.initializeContext(DefaultApplicationContext.java:249)
	at app//io.micronaut.context.DefaultBeanContext.readAllBeanDefinitionClasses(DefaultBeanContext.java:3326)
	at app//io.micronaut.context.DefaultBeanContext.finalizeConfiguration(DefaultBeanContext.java:3684)
	at app//io.micronaut.context.DefaultBeanContext.start(DefaultBeanContext.java:341)
```. 

We will not be able to write a function test for this feature if the generated configuration is invalid.

**Review by sdelamo — CHANGES_REQUESTED** [review]
I have created a PR with suggested changes: 

https://github.com/micronaut-projects/micronaut-starter/pull/1589

File: starter-core/src/main/java/io/micronaut/starter/feature/database/DataAzureCosmosFeature.java (hunk 1 above)
Comment (radovanradic) [reply starter-core/src/main/java/io/micronaut/starter/feature/database/DataAzureCosmosFeature.java]: Don't know if it makes sense then without test resources support, there will never be option to have valid placeholder here in config. We have this class that triggers creation of cosmos client etc.
```
@Requires(classes = CosmosClient.class)
final class CosmosDatabaseInitializer {

 @PostConstruct
    void initialize(CosmosClient cosmosClient,
                    RuntimeEntityRegistry runtimeEntityRegistry,
                    @Nullable
                    CosmosDiagnosticsProcessor cosmosDiagnosticsProcessor,
                    CosmosDatabaseConfiguration configuration) {
```
Maybe adding 
```
@Requires(property = "azure.cosmos.endpoint")
```
would make sense and then we can create this feature without endpoint and key.

Comment (sdelamo) [reply starter-core/src/main/java/io/micronaut/starter/feature/database/DataAzureCosmosFeature.java]: Agree.  if `CosmosDatabaseInitializer` requires  a property to be present it should be added it as an `@Requires`. Maybe something like: 

```java
@Requires(bean= CosmosDatabaseConfiguration.class, beanProperty="endpoint")
```

radovanradic: There isn't such field on `CosmosDatabaseConfiguration` so this is done via 
```
@ConfigurationBuilder(prefixes = "")
 protected CosmosClientBuilder cosmosClientBuilder = new CosmosClientBuilder();
```
so `CosmosClientBuilder` (from azure cosmos lib) gets endpoint and key properties. Don't know how to configure `@Requires` in that case but tried this `@Requires(property = "azure.cosmos.endpoint")` and it worked

starter-core/src/test/groovy/io/micronaut/starter/feature/database/DataAzureCosmosFeatureSpec.groovy
```diff
@@ -0,0 +1,83 @@
...
+        feature.category == Category.DATABASE
+        feature.name == 'data-azure-cosmos'
+    }
+
+    void "test maven data-azure-cosmos feature adds dependency"() {
+        when:
+        String template = new BuildBuilder(beanContext, BuildTool.MAVEN)
+                .features(["data-azure-cosmos"])
+                .render()
+        template.count('''
```
sdelamo: Please, use the `BuildTestVerifier` API introduce here: https://github.com/micronaut-projects/micronaut-starter/pull/1590

starter-core/src/test/groovy/io/micronaut/starter/feature/database/DataAzureCosmosFeatureSpec.groovy
```diff
@@ -0,0 +1,83 @@
...
+        def pom = new XmlSlurper().parseText(template)
+
+        then:
+        with(pom.dependencies.dependency.find { it.artifactId == "micronaut-data-azure-cosmos" }) {
+            scope.text() == 'compile'
+            groupId.text() == 'io.micronaut.data'
+        }
+    }
+
+    void "test gradle data-azure-cosmos feature adds dependency"() {
```
sdelamo: with the `BuildTestVerifier`Api you can remove this test and test every build combination in the previous test. see: https://github.com/micronaut-projects/micronaut-starter/pull/1590

starter-core/src/main/java/io/micronaut/starter/feature/database/DataAzureCosmosFeature.java
radovanradic: I have merged this PR https://github.com/micronaut-projects/micronaut-data/pull/1968 and when next version is released (should be 3.9.5 but don't know where should it be updated also so micronaut-starter sees that version) then we can finish this feature work.

radovanradic: I have updated default configuration to not contain endpoint and key since makes no sense. Changes are done also in micronaut-data:azure-data-cosmos to not initialize when endpoint and key are not set. That is in version 3.9.5 now.
But don't know what else is needed to micronaut-starter to pickup correct version as now it generates this in maven for example 
```
<micronaut.data.version>3.9.4</micronaut.data.version>
```

radovanradic:
@sdelamo I have updated to the latest micronaut version (actually to 3.9.0-SNAPSHOT same as in 3.9.x branch) and all of sudden all checks are failing. Do you know what went wrong?

radovanradic:
After changing plugin to `id 'io.micronaut.build.shared.settings' version '5.3.18'` then most builds are passing but there are still some aws tests failing which is likely unrelated to these changes here.

sdelamo (approved):

starter-core/src/test/groovy/io/micronaut/starter/feature/database/DataAzureCosmosFeatureSpec.groovy
radovanradic: Removed, thanks.

sdelamo: thanks @radovanradic

PR #1585, opened 2023-01-12
Fix R2dbc ordering and integration with TestResources

timyates:
We had intermittent failures as R2dbc and DataR2dbc had the same ordering so the result was non-repeatable.

This PR applies the correct ordering.

We also did not handle Test Resources properly, we handled it for HibernateReactive, but R2DBC was missed.


sdelamo (approved):

timyates:
Need to fix maven...

sdelamo (approved):

PR #1592, opened 2023-01-16
Add gradle-enterprise feature to Maven builds

### LINKED ISSUE #1517: GradleEnterprise feature should be enabled for Maven builds
Currently the Gradle Enterprise feature only works for Gradle builds

We should support Maven builds also

alextu:
This solves https://github.com/micronaut-projects/micronaut-starter/issues/1517

When adding `gradle-enterprise` feature to a Maven project, it will generate:
- a `.mvn/extensions.xml` with Gradle Enterprise and Common Custom User Data Maven extensions defined
- a `.mvn/gradle-enterprise.xml` with no server defined by default, accepting TOS so it can push to scans.gradle.com
This PR also refactors the `micronaut-gradle-enterprise` feature to extract the common generation logic. It also removes the `buildCache.remote.server.credentials` attribute from `gradle-enterprise.xml` since it's not required [anymore](https://gradle.com/enterprise/releases/2022.3/#manage-build-cache-access-control-via-gradle-enterprise-identity-management) when using the builtin cache node.


alvarosanchez:
Can you please add the XML namespace to the `extensions.xml` and `gradle-enterprise.xml` files? You have examples here: 

https://github.com/micronaut-projects/micronaut-maven-plugin/tree/master/.mvn

alvarosanchez (approved):

starter-core/src/test/groovy/io/micronaut/starter/feature/build/gradle/GradleEnterpriseSpec.groovy
```diff
@@ -42,4 +44,45 @@ class GradleEnterpriseSpec extends ApplicationContextSpec implements CommandOutp
...
         buildTool << [BuildTool.GRADLE, BuildTool.GRADLE_KOTLIN]
     }
+
+    void 'feature gradle-enterprise creates a .mvn/extensions dot xml file'() {
+        when:
+        Map<String, String> output = generate(ApplicationType.DEFAULT, new Options(Language.JAVA, BuildTool.MAVEN), ["gradle-enterprise"])
+        def xml = new XmlParser().parseText(output[".mvn/extensions.xml"])
+
+        then:
+        xml.name() == 'extensions'
```
timyates:
```suggestion
        xml.name().localPart == 'extensions'
```

starter-core/src/test/groovy/io/micronaut/starter/feature/build/gradle/GradleEnterpriseSpec.groovy
```diff
@@ -42,4 +44,45 @@ class GradleEnterpriseSpec extends ApplicationContextSpec implements CommandOutp
...
+        buildTool << [BuildTool.GRADLE, BuildTool.GRADLE_KOTLIN]
+    }
+
+    void 'feature gradle-enterprise creates a .mvn/gradle-enterprise dot xml file'() {
+        when:
+        Map<String, String> output = generate(ApplicationType.DEFAULT, new Options(Language.JAVA, BuildTool.MAVEN), ["gradle-enterprise"])
+        def xml = new XmlParser().parseText(output[".mvn/gradle-enterprise.xml"])
+
+        then:
+        xml.name() == 'gradleEnterprise'
```
timyates:
```suggestion
        xml.name().localPart == 'gradleEnterprise'
```

starter-core/src/test/groovy/io/micronaut/starter/feature/build/gradle/GradleEnterpriseSpec.groovy
```diff
@@ -42,4 +44,45 @@ class GradleEnterpriseSpec extends ApplicationContextSpec implements CommandOutp
...
+        !xml.server
+        xml.buildScan.termsOfService.url.text() == 'https://gradle.com/terms-of-service'
+        xml.buildScan.termsOfService.accept.text() == 'yes'
```
timyates: This fails, as it is `'true'`, not `'yes'` 🤔

timyates (changes requested):

starter-core/src/test/groovy/io/micronaut/starter/feature/build/gradle/GradleEnterpriseSpec.groovy
alextu: Yeah it's not consistent with the Gradle version, I've fixed it by having `GradleEnterpriseConfiguration#getTermsOfServiceAgree` returning a boolean and handling differently in templates

starter-core/src/test/groovy/io/micronaut/starter/feature/build/gradle/GradleEnterpriseSpec.groovy
alextu: I used instead `xml.name() == new QName('https://www.gradle.com/gradle-enterprise-maven', 'gradleEnterprise')`

timyates (approved):

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/GradleEnterpriseConfiguration.java
```diff
@@ -34,7 +34,7 @@ default String getTermsOfServiceUrl() {
     }
 
     @Nullable
-    default String getTermsOfServiceAgree() {
-        return "yes";
+    default Boolean getTermsOfServiceAgree() {
```
sdelamo, line 37: this is a breaking change.

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/MavenGradleEnterprise.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package io.micronaut.starter.feature.build.gradle;
+
+import com.fizzed.rocker.RockerModel;
+import io.micronaut.starter.application.generator.GeneratorContext;
+import io.micronaut.starter.feature.build.maven.templates.extensions;
+import io.micronaut.starter.template.RockerTemplate;
+import jakarta.inject.Singleton;
+
+@Singleton
```
sdelamo, line 24: I would prefer to have this logic inside the `GradleEnterprise` feature with protected methods so that they could be overriden instead of having an extra `@Singleton`

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/gradleEnterprise.rocker.raw
```diff
@@ -1,19 +1,26 @@
...
+            <url>@(config.getTermsOfServiceUrl())</url>
+        }
+        @if (config.getTermsOfServiceAgree() != null && config.getTermsOfServiceAgree()) {
+            <accept>true</accept>
+        }
+        </termsOfService>
+    }
         <publish>ALWAYS</publish>
     </buildScan>
-    <buildCache>
```
sdelamo, line 9: I think this `buildCache` configuration was used via the starter build with the `MicronautGradleEnterprise` feature.  

I don't think we should remove them.

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/GradleEnterpriseConfiguration.java
sdelamo: I have created a PR on top of yours deprecating this method instead of removing it https://github.com/micronaut-projects/micronaut-starter/pull/1595/commits/6f356eb6b5915b46c9ab372060558bf8a634b219

are you ok wit this change @alextu ?

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/MavenGradleEnterprise.java
sdelamo: I have removed `MavenGradleEnterprise`here: 

https://github.com/micronaut-projects/micronaut-starter/pull/1595/commits/c9b28bc5df6d7688dad2cf9e65e189832d2e7be3

@alextu let me know if you are ok with this change?

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/gradleEnterprise.rocker.raw
sdelamo: I have added the buildCache back here: 
https://github.com/micronaut-projects/micronaut-starter/pull/1595/commits/7064a1daa2606d00eead912f81dc8ca0b2a2b8a5

@alextu are you ok with this change?

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/GradleEnterpriseConfiguration.java
alextu: Sure, no problem

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/MavenGradleEnterprise.java
alextu: Sure, no problem

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/gradleEnterprise.rocker.raw
alextu: So my intention was to adapt the snippet to the latest GE components: 

> When using Maven extension version 1.15+, Gradle Enterprise 2022.3+ and Gradle Enterprise Build Cache Nodes 13+, the same access key credential used for Build Scan publishing, Test Distribution and other functions can be used for build cache access control.

And since the build cache is enabled by default, there's no need to add a `buildCache` element at all. 
That being said I'm not familiar with how is used this `MicronautGradleEnterprise` feature, feel free to keep it if you think it's still required

sdelamo: @alextu this is how we use it: 

https://github.com/micronaut-projects/micronaut-starter/blob/3.8.x/test-utils/src/main/groovy/io/micronaut/starter/test/CommandSpec.groovy#L124-L127

We add this hidden feature to the maven applications we generate as part of the functional test suite. The goal of this is to reduce the test execution time. 

thoughts @timyates @alvarosanchez ?

alextu: To clarify a bit: with this change, given the build has the GE access key provisionned, building with the remote build cache should work as it currently does with the credentials specified. If it's not the case:
- passing the access key would be best (and checking the build cache read permissions are set for this key)
- if not possible, I would parametrized this template so setting the build cache credentials is the exception for this particular use case, not the norm

Does that make sense ?

sdelamo: Do you think is not necessary? 

```xml
    <buildCache>
         <remote>
             <server>
                 <credentials>
                     <username>${env.GRADLE_ENTERPRISE_CACHE_USERNAME}</username>
                     <password>${env.GRADLE_ENTERPRISE_CACHE_PASSWORD}</password>
                 </credentials>
             </server>
         </remote>
     </buildCache>
```

Thus, the maven applications generated in CI will use the build cache through our gradle enterprise instance automatically because the environment variables are present in CI. 

Is that correct?

alextu: Exactly, note also that the provisionned Gradle enterprise access key associated user should also have "Read build cache data" and "Write build cache data" permissions. Best would be to check such a build on this branch, could you point me to one ?

sdelamo:
@alextu I have merged https://github.com/micronaut-projects/micronaut-starter/pull/1595 which contains your changes. Let me know if you have any concerns

PR #1613, opened 2023-02-15
Upgrade to Gradle plugin 3.7.2

melix:


sdelamo:
@melix do you know about the error?

PR #1618, opened 2023-02-18
ci: use 22.3.1 GraalVM and java 17

sdelamo:


sdelamo:
it fails with: 

```
Fatal error: java.lang.IllegalAccessError: class io.micronaut.buffer.netty.NettyFeature 
(in unnamed module @0x51d50b5f) cannot access class 
com.oracle.svm.core.jdk.SystemPropertiesSupport (in module org.graalvm.nativeimage.builder) 
because module org.graalvm.nativeimage.builder does not export 
com.oracle.svm.core.jdk to unnamed module @0x51d50b5f
```

@micronaut-projects/core-developers any idea?

melix:
You are basically facing the issue that the Maven and Gradle plugins have when building Micronaut applications, [for which we implemented a workaround](https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/e40fe5f4fd4b27602e14156c344928c3b93f7a77/graalvm-plugin/src/main/java/io/micronaut/gradle/graalvm/MicronautGraalPlugin.java#L186-L190) in the plugins.

So basically you need to do the same. Alternatively I guess `cli` would need to be updated to be built using the Micronaut Gradle plugin, which would take care of this.

graemerocher:
@sdelamo if you approve and merge https://github.com/micronaut-projects/micronaut-core/pull/8821 then this issue will probably be fixed without additional flags

sdelamo:
> @sdelamo if you approve and merge [micronaut-projects/micronaut-core#8821](https://github.com/micronaut-projects/micronaut-core/pull/8821) then this issue will probably be fixed without additional flags

We won't be able to merge this until 4.0.x then?

PR #1636, opened 2023-03-01
Rework how the dependency coordinates found in templates are handled

### LINKED ISSUE #1635: Consider reworking internal `pom.xml`
We use a [checked-in pom.xml](https://github.com/micronaut-projects/micronaut-starter/blob/be9d26ab723d39010b483822daed91d1d88bf1ec/starter-core/src/main/resources/pom.xml) file to handle dependency coordinates of dependencies which are added by the project generator.

This file is static, and uses the `pom.xml` format so that dependabot can update dependencies automatically.

In practice, it is quite cumbersome to use. Especially, any Gradle plugin update requires modifying [3 files](https://github.com/micronaut-projects/micronaut-starter/pull/1631) and, as far as I understand, we can't even use properties in that pom because they would not be recognized by [our custom parsing logic](https://github.com/micronaut-projects/micronaut-starter/blob/414d81590cbc80d76d6b111e7273bd7eda8695f8/starter-core/src/main/java/io/micronaut/starter/build/dependencies/CoordinatesUtils.java#L38).

Instead, I suggest we use a similar approach to what we have in other projects, which is to use a version catalog, updated via Renovatebot, which is then used to generate code constants.

melix:
Instead of using a pom.xml file, this now makes use of a Gradle version catalog, found in `gradle/templates.versions.toml`. This file should be handled by renovatebot automatically.

This will typically make the Gradle plugin upgrades much easier to handle since there will be a single place to change.

Fixes #1635


timyates (approved):
🧙

timyates:
I believe there's a fix in the pipeline from @sdelamo for the aws/groovy failures

melix:
This could be updated to Gradle plugin 3.7.4 but given the very slow feedback loop, I won't push this today.

starter-core/src/test/groovy/io/micronaut/starter/feature/dependencies/MultipleSourcesDependencyResolver.groovy
```diff
@@ -4,16 +4,20 @@ import io.micronaut.context.annotation.Replaces
...
 import io.micronaut.starter.build.dependencies.Coordinate
 import io.micronaut.starter.build.dependencies.CoordinatesUtils
+import io.micronaut.starter.build.dependencies.PomCoordinatesUtils
 import io.micronaut.starter.build.dependencies.PomDependencyVersionResolver
 import jakarta.inject.Singleton
 
 @Replaces(PomDependencyVersionResolver.class)
 @Singleton
-class MultiplePomDependencyResolver implements PomDependencyVersionResolver {
+class MultipleSourcesDependencyResolver implements PomDependencyVersionResolver {
```
sdelamo, line 13: please, keep the old name to avoid an unnecessary breaking change.

starter-core/src/test/groovy/io/micronaut/starter/build/dependencies/PomCoordinatesUtils.java
```diff
@@ -35,12 +35,12 @@
...
 
-public final class CoordinatesUtils {
-    private static final Logger LOG = LoggerFactory.getLogger(CoordinatesUtils.class);
+public final class PomCoordinatesUtils {
+    private static final Logger LOG = LoggerFactory.getLogger(PomCoordinatesUtils.class);
 
     private static final String NODE_NAME_TEXT = "#text";
 
-    private CoordinatesUtils() {
+    private PomCoordinatesUtils() {
```
sdelamo: please, keep the old name to avoid an unnecessary breaking change.

sdelamo (changes requested):
Please, note that we also use the pom.xml in guides. https://github.com/micronaut-projects/micronaut-guides/blob/master/buildSrc/src/main/resources/pom.xml

I assume this is compatible to have another version catalogue in guides.

starter-core/src/test/groovy/io/micronaut/starter/feature/dependencies/MultipleSourcesDependencyResolver.groovy
melix: I don't understand what breaking change you're talking about : this is a test.

starter-core/src/test/groovy/io/micronaut/starter/build/dependencies/PomCoordinatesUtils.java
melix: Same question here : this is a test, how can it be a breaking change?

melix:
> we also use the pom.xml in guides. https://github.com/micronaut-projects/micronaut-guides/blob/master/buildSrc/src/main/resources/pom.xml

Oh, so you mean the `CoordinatesUtils` class is used in guides too? Then you're right I need to make changes to keep this working.

starter-core/src/test/groovy/io/micronaut/starter/build/dependencies/PomCoordinatesUtils.java
melix: Forget what I said, it's in `test` because that's where I moved the class.

starter-core/src/test/groovy/io/micronaut/starter/feature/dependencies/MultipleSourcesDependencyResolver.groovy
sdelamo: I did not realise was a test. Sorry for the noise.

sdelamo (approved):

PR #1651, opened 2023-03-09
implement support for toml array tables

yawkat:
Also has some build changes


melix (approved):

sdelamo (changes requested):
This PR should target 3.9.x

yawkat:
@sdelamo it's not required for starter (we don't really produce arrays) and i want to use records

sdelamo:
> @sdelamo it's not required for starter (we don't really produce arrays) and i want to use records

why are we adding it to starter if we don't need it?

yawkat:
so that we don't have diverging code between starter and build

sdelamo:
@yawkat closing this for now.

PR #1659, opened 2023-03-15
support api gradle configuration

sdelamo:


timyates (approved):
I added a test that Maven remains the same

PR #1661, opened 2023-03-15
MicronautDependencyUtils gcp

sdelamo:


starter-core/src/main/java/io/micronaut/starter/build/dependencies/MicronautDependencyUtils.java
```diff
@@ -18,7 +18,6 @@
 import io.micronaut.core.annotation.NonNull;
 
 public final class MicronautDependencyUtils {
-
```
timyates, line 21: I kinda like the blank line after the class definition 🤷‍♂️

timyates (approved):

PR #1668, opened 2023-03-15
extract HandlerProvider [ci skip]

sdelamo:


starter-core/src/main/java/io/micronaut/starter/feature/function/template/handlerReadme.rocker.raw
```diff
@@ -5,13 +5,13 @@
...
 @args (
-Feature feature,
+HandlerClassFeature feature,
 ApplicationType applicationType,
 Project project,
 DocumentationLink documentationLink
 )
 
-@if (feature instanceof HandlerClassFeature && ((HandlerClassFeature) feature).handlerClass(applicationType, project) != null) {
+@if (feature.handlerClass(applicationType, project) != null) {
```
timyates: This is never null at the moment...  All instances of `HandlerClassFeature` have an implementation that cannot return null, however it's marked as `Nullable` so I guess there could be future implementations where it is null...

timyates (approved):

starter-core/src/main/java/io/micronaut/starter/feature/function/template/handlerReadme.rocker.raw
sdelamo: remove if for handler https://github.com/micronaut-projects/micronaut-starter/pull/1668/commits/f2a32d0c9fbc5a01e4a822e546dd2308677916a5

PR #1670, opened 2023-03-15
Gradle repository mavenContent filters

timyates:
We had these for Maven, but not Gradle.  This brings the two into line


timyates:
So for a kotlin gradle build, a snapshot repo will be written as:

```
    maven("https://s01.oss.sonatype.org/content/repositories/snapshots/") {
        mavenContent { snapshotsOnly() }
    }
```

Instead of the plain

```
    maven("https://s01.oss.sonatype.org/content/repositories/snapshots/")
```

And for a Groovy DSL build we will have

```
    maven { url "https://s01.oss.sonatype.org/content/repositories/snapshots/"
        mavenContent { snapshotsOnly() }
    }
```

instead of

```
    maven { url "https://s01.oss.sonatype.org/content/repositories/snapshots/" }
```

PR #1672, opened 2023-03-16
Switch default JDK in Micronaut Launch to 17

timyates:
This leaves the default as 11 when creating a lambda (as AWS don't yet fully support 17)


graemerocher (approved):

timyates:
I suspect we see low java 17 figures in micronaut launch as 11 is the default, and people stick with that

sdelamo (approved):

PR #1673, opened 2023-03-17
Show the available JDK options in the CLI --help

timyates:
I was going to go further, and show the default, but the default for AWS is 11 (depending on service). Also, the default language changes the default gradle dsl and the default test framework. This means I don't believe there's much more we can do of value.


sdelamo:
@timyates ci fails

sdelamo (approved):

timyates:
Should be fixed now, sorry about that

sdelamo (approved):

PR #1689, opened 2023-04-03
Add configuration to camunda-platform7 and update metadata of Camunda related projects

tobiasschaefer:
I'm the lead developer of the Micronaut Camunda Integration Project.

Please merge this PR which adds configuration options and makes the setup easier.

I tested the following:
./gradlew micronaut-cli:run --args="create-app --list-features" -> Looks good, feature is listed

Both the following look good
./gradlew micronaut-cli:run --args="create-app --features=camunda-platform7 temp-gradle"
./gradlew micronaut-cli:run --args="create-app --build=maven --features=camunda-platform7 temp-maven"

and build successfully.

FYI: @sdelamo

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/camunda/Platform7.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/camunda/Platform7.java
@@ -23,17 +23,30 @@ import io.micronaut.starter.build.dependencies.Dependency;
 import io.micronaut.starter.feature.Category;
 import io.micronaut.starter.feature.FeatureContext;
 import io.micronaut.starter.feature.database.DatabaseDriverFeature;
+import io.micronaut.starter.feature.server.Jetty;
+import io.micronaut.starter.feature.server.Netty;
 import io.micronaut.starter.feature.test.AssertJ;
 import jakarta.inject.Singleton;
 
 @Singleton
 public class Platform7 implements CamundaCommunityFeature {
+    public static final String NAME = "camunda-platform7";
 
-    private final AssertJ assertJ;
+    private static final Dependency.Builder DEPENDENCY_PLATFORM7 = Dependency.builder()
+            .lookupArtifactId("micronaut-camunda-bpm-feature")
+            .compile();
+
+    private static final Dependency.Builder DEPENDENCY_BPM_ASSERT = Dependency.builder()
+            .lookupArtifactId("camunda-bpm-assert")
+            .test();
+
     private final DatabaseDriverFeature defaultDbFeature;
+    private final Jetty jetty;
+    private final AssertJ assertJ;
 
-    public Platform7(DatabaseDriverFeature defaultDbFeature, AssertJ assertJ) {
+    public Platform7(DatabaseDriverFeature defaultDbFeature, Jetty jetty, AssertJ assertJ) {
         this.defaultDbFeature = defaultDbFeature;
+        this.jetty = jetty;
         this.assertJ = assertJ;
     }
 
@@ -46,7 +59,7 @@ public class Platform7 implements CamundaCommunityFeature {
     @Override
     @NonNull
     public String getCommunityFeatureTitle() {
-        return "Camunda Workflow Engine";
+        return "Camunda Platform 7 Workflow Engine";
     }
 
     @Override
@@ -56,7 +69,7 @@ public class Platform7 implements CamundaCommunityFeature {
 
     @Override
     public String getDescription() {
-        return "Bringing process automation to Micronaut: Embed the Camunda Platform Workflow Engine";
+        return "Bringing process automation to Micronaut: Embed the Camunda Platform 7 Workflow Engine";
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/camunda/Platform7.java
```diff
@@ -23,17 +23,21 @@
...
 @Singleton
 public class Platform7 implements CamundaCommunityFeature {
 
-    private final AssertJ assertJ;
     private final DatabaseDriverFeature defaultDbFeature;
+    private final Jetty jetty;
+    private final AssertJ assertJ;
 
-    public Platform7(DatabaseDriverFeature defaultDbFeature, AssertJ assertJ) {
+    public Platform7(DatabaseDriverFeature defaultDbFeature, Jetty jetty, AssertJ assertJ) {
```
sdelamo, line 47: Annotate the new constructor with @Inject. Keep the old constructor deprecated.

starter-core/src/test/groovy/io/micronaut/starter/feature/camunda/Platform7Spec.groovy
```diff
@@ -83,8 +85,28 @@ class Platform7Spec extends ApplicationContextSpec implements CommandOutputFixtu
       <artifactId>h2</artifactId>
       <scope>runtime</scope>
     </dependency>
+''') == 1
+        template.count('''\
```
sdelamo: Please, use the new BuildVerifierApi to test dependencies.

starter-core/src/main/java/io/micronaut/starter/feature/camunda/Platform7.java
```diff
@@ -66,16 +70,26 @@ public boolean supports(ApplicationType applicationType) {
...
+            featureContext.addFeature(jetty );
+        }
         if (!featureContext.isPresent(AssertJ.class)) {
             featureContext.addFeature(assertJ);
         }
     }
 
     @Override
     public void apply(GeneratorContext generatorContext) {
+        generatorContext.getConfiguration().put("camunda.admin-user.id", "admin");
```
sdelamo: extract `protected void addConfiguration(GeneratorContext)` method with all these statements.

starter-core/src/main/java/io/micronaut/starter/feature/camunda/Platform7.java
```diff
@@ -66,16 +70,26 @@ public boolean supports(ApplicationType applicationType) {
 
     @Override
     public void processSelectedFeatures(FeatureContext featureContext) {
+        featureContext.exclude(feature -> feature instanceof Netty);
         if (!featureContext.isPresent(DatabaseDriverFeature.class)) {
             featureContext.addFeature(defaultDbFeature);
         }
+        if (!featureContext.isPresent(Jetty.class)) {
```
sdelamo: use `FeatureContext::addFeatureIfNotPresent`

sdelamo (changes requested):

starter-core/src/test/groovy/io/micronaut/starter/feature/camunda/Platform7Spec.groovy
tobiasschaefer: Thanks for the hint.
I switched to BuildTestVerifier (I guess that is what you are referring to)

starter-core/src/main/java/io/micronaut/starter/feature/camunda/Platform7.java
tobiasschaefer: I don't see why we need to keep the old constructor. IMHO it would never be called. We need the new constructor to initialize all fields.

tobiasschaefer:
Hi @sdelamo ,

thank you for your comments in the PR.

I updated the PR to consider your comments.

Note: I updated `GradleBuildTestVerifier` to match the behaviour of `MavenBuildTestVerifier`, i.e. it ignores the dependency version information if the dependency is not managed. This is needed because all Camunda related dependencies are currently not managed by the micronaut-bom. If you give me hints how to accomplish this I'm happy to do it in a separate PR.

sdelamo (approved):

wetted:
> Note: I updated `GradleBuildTestVerifier` to match the behaviour of `MavenBuildTestVerifier`, i.e. it ignores the dependency version information if the dependency is not managed. This is needed because all Camunda related dependencies are currently not managed by the micronaut-bom. If you give me hints how to accomplish this I'm happy to do it in a separate PR.

@tobiasschaefer @sdelamo 
This change breaks `GradleBuildTestVerifier` and needs to be fixed. For example, if I have the following,

```
        verifier.hasDependency("io.micronaut.oraclecloud", "micronaut-oraclecloud-function-http", Scope.COMPILE)
        !verifier.hasDependency("io.micronaut.oraclecloud", "micronaut-oraclecloud-function")
```

The second check fails because the check for "micronaut-oraclecloud-function" is matching "micronaut-oraclecloud-function-http". The regex is flawed in that regard. I'll work on a fix for it in my PR. The motivation is just to ignore version notation in the dependency, right?

tobiasschaefer:
Hi @wetted ,

yes, the motivation is just to ignore version notation in the dependency (so that the Gradle Implementation matches the Maven Implementation).

Thanks for fixing the issue for the oracle-function-http feature 🥇 Strange though, that all tests were green.

wetted:
@tobiasschaefer the tests were green for you. I added a new test that exposed the need for a fix. :)

PR #1700, opened 2023-04-11
AWS-lambda features without GraalVM feature raises validation error when Java 17 is selected

### LINKED ISSUE #1699: AWS-lambda without Graalvm should throw an error if Java 17 is selected
Currently, Lambda does not have a 17 runtime.

wetted:
closes #1699

This PR validates that AWS Lambda features cannot be selected for JDK 17, unless the GraalVM featre is also selected, since AWS doesn't yet have a a runtime. 

However, this is a bit more invloved than just changing the feature and adding new tests for the validator. I had to disable tests for JDK 17 that use AWS Lambda features without Graalvm feature, a combination no longer allowed with `AwsLambdaFeatureValidator`.

But this is rather a bit of smell. Some of these tests are not directly related to testing `AwsLambda` features, but merely test something else for all application types, where `ApplicationType.FUNCTION` by default selects the "aws-lambda" feature, which in turn now results in raising an  exception for JDK 17. With Micronaut 3.x these tests will run when CI is JDK 8 and JDK 11. However, with Micronaut 4 they won't run at all. Maybe an alternative is to rewrite those tests to either include the "graalvm" feature or a JDK 17-supported FUNCTION feature (OCI or Azure).

It also seems to me that for Micronaut 4, Creating a "Function Application for Serverless" should no longer apply the "aws-lambda" feature by default, or if it does it needs to include the "graalvm' feature to work. Maybe this is the fix for this situation?

@sdelamo thoughts?

### DIFF
--- a/starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateLambdaBuilderCommand.java
+++ b/starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateLambdaBuilderCommand.java
@@ -27,6 +27,7 @@ import io.micronaut.starter.feature.architecture.CpuArchitecture;
 import io.micronaut.starter.feature.architecture.X86;
 import io.micronaut.starter.feature.aws.AmazonApiGateway;
 import io.micronaut.starter.feature.aws.AwsApiFeature;
+import io.micronaut.starter.feature.aws.AwsLambdaFeatureValidator;
 import io.micronaut.starter.feature.aws.Cdk;
 import io.micronaut.starter.feature.aws.LambdaFunctionUrl;
 import io.micronaut.starter.feature.aws.LambdaTrigger;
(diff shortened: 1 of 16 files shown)


wetted:
The functional/integration tests need work still, for the same reasons as above.

sdelamo:
> But this is rather a bit of smell. Some of these tests are not directly related to testing `AwsLambda` features, but merely test something else for all application types, where `ApplicationType.FUNCTION` by default selects the "aws-lambda" feature, which in turn now results in raising an  exception for JDK 17. With Micronaut 3.x these tests will run when CI is JDK 8 and JDK 11. However, with Micronaut 4 they won't run at all. Maybe an alternative is to rewrite those tests to either include the "graalvm" feature or a JDK 17-supported FUNCTION feature (OCI or Azure).

AWS lambda will have a 17 runtime in time for Micronaut 4. 

I have changed your tests to remove `@IgnoreIf`

wetted:
> I have changed your tests to remove `@IgnoreIf`

OK. Thanks. What you changed to is better. I'll work on the broken builds (tests).

sdelamo:
> . I'll work on the broken builds (tests).

I think this should fix the JDK 8 build https://github.com/micronaut-projects/micronaut-starter/pull/1700/commits/134bed980680a637d7d960d518e937ee9fdb3f21

PR #1701, opened 2023-04-12
pin baseImage to amazonlinux:2

sdelamo:
see: https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/691


graemerocher:
we will need to remember to update this if the plugin is upgraded

graemerocher (approved):

sdelamo:
> we will need to remember to update this if the plugin is upgraded

I have a bash script to test this https://github.com/micronaut-projects/micronaut-starter/blob/3.8.x/test-lambda-graalvm.sh

But I was running it only between minor versions. 

I will run it in every patch version as well.

PR #1704, opened 2023-04-12
Add KSP feature

graemerocher:
Adds a feature for KSP. Not fully tested yet as 4.0.x branch is broken, but ran tests against 3.9.x branch and they worked.


starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleConfiguration.java
```diff
@@ -136,4 +149,18 @@ public static Optional<GradleConfiguration> of(@NonNull Scope scope,
         }
         return Optional.empty();
     }
+
+    private static GradleConfiguration kotlinAnnotationProcessor(GeneratorContext generatorContext) {
```
sdelamo, line 156: We need a test

sdelamo (approved):

Issue #1698, opened 2023-04-11
Micronaut Validation should add an annotation processor and a dependency


sdelamo:
https://micronaut-projects.github.io/micronaut-validation/snapshot/guide/#quickStart

New Group id is `io.micronaut.validation`

For maven, as we do with Micronaut Data, we will need to set a property `micronaut.validation.version`. 

ping @alvarosanchez

sdelamo:
@wetted To start, We should move the logic of adding the dependencies outside of rocker into Java classes. If you want create such a PR and target 3.9.x

alvarosanchez:
We don't need to set the `micronaut.validation.version` property in user POMs, it's inherited from the Platform BOM. The same applies to `micronaut.data.version`, don't know why it's there

PR #1707, opened 2023-04-13
Refactor rocker dependencies into java Feature API (partial)

### LINKED ISSUE #1681: Generated projects with `google-cloud-function-http` feature include no-longer-needed dependencies
The project template generated when a project includes the `google-cloud-function-http` feature includes the following comment:

https://github.com/micronaut-projects/micronaut-starter/blob/5a81642955e30defcb5f7c602f3b93e4fd1e7a1b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw#L104

https://github.com/GoogleCloudPlatform/functions-framework-java/pull/32 has long since been merged and I think the additional dependencies can be removed. I'd be happy to open a pull request, but it's not entirely clear to me which of the dependencies (just the dependency following the comment? all of the dependencies following the comment?) are affected.

wetted:
On Issue #1698 @sdelamo says, "we should move the logic of adding the dependencies outside of rocker into Java classes."

This implements a large portion of that. What's still left is somewhat more complicated – with conditional logic between features, inconsistencies between gradle and maven, and for other reasons – but there's no reason that should hold this part up.

The dependency logic for the azure features is broken for gradle. That's a bug so I will fix that separately.


sdelamo:
@wetted is this ready for review? it is still in draft

wetted:
> @wetted is this ready for review? it is still in draft

@sdelamo no, the unit tests pass, but something is broken in the functional/integration tests I need to address.

wetted:
I am working the Azure, GCP, and OracleCloud Rocker template refactors in separate PRs. They are somewhat more complicated, and have bugs for Gradle projects.

wetted:
> @wetted is this ready for review? it is still in draft

@sdelamo this is good to go now.

starter-core/src/main/java/io/micronaut/starter/feature/function/awslambda/AwsLambda.java
```diff
@@ -246,6 +251,12 @@ private void addDependencies(@NonNull GeneratorContext generatorContext) {
         if (generatorContext.hasFeature(AwsLambdaSnapstart.class)) {
             generatorContext.addDependency(DEPENDENCY_MICRONAUT_CRAC);
         }
+
+        if (generatorContext.getFeatures().testFramework().isSpock() &&
+                generatorContext.getBuildTool().isGradle()) {
+            // maven has this in parent pom
+            generatorContext.addDependency(DEPENDENCY_MICRONAUT_FUNCTION_TEST);
+        }
```
alvarosanchez, line 259: Would it make sense to move this to the parent `AwsCloudFeature`?

starter-core/src/main/java/io/micronaut/starter/feature/function/gcp/GoogleCloudFunction.java
```diff
@@ -63,6 +64,11 @@ public boolean isVisible() {
         return false;
     }
 
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        super.apply(generatorContext);
+    }
```
alvarosanchez: Is this needed?

starter-core/src/main/java/io/micronaut/starter/feature/other/OpenApi.java
```diff
@@ -56,6 +58,14 @@ public void apply(GeneratorContext generatorContext) {
                 .groupId("io.swagger.core.v3")
                 .artifactId("swagger-annotations")
                 .compile());
+        if (generatorContext.getBuildTool() == BuildTool.MAVEN &&
+                generatorContext.getLanguage() == Language.GROOVY) {
+            generatorContext.addDependency(Dependency.builder()
+                    .groupId("io.micronaut.openapi")
+                    .artifactId("micronaut-openapi")
+                    .versionProperty("micronaut.openapi.version")
```
alvarosanchez: This is probably in the BOM, so the version should be needed

alvarosanchez (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/function/awslambda/AwsLambda.java
wetted: @alvarosanchez I don't think we want to put it there. `AwsCloudFeature` is an interface that extends `CloudFeature` and not `Feature`. What I have now still matches the logic that was in the Rocker template.

wetted:
@alvarosanchez @sdelamo 

I added another change set that refactors the GCP dependencies from rocker files into the Feature classes. Please review that as well.

alvarosanchez (approved):

alvarosanchez (approved):

wetted:
I am now able to duplicate the failing test locally. I am working on fixing it.

PR #1709, opened 2023-04-14
Fix missing dependencies for Gradle projects with "oracle-function-http" feature,

wetted:
Fixing missing dependencies for Gradle projects with "oracle-function-http" feature, to make consistent with Maven projects. Refactor dependency logic from Rocker files to Feature classes.

This also fixes a bug from a change to `GradleBuildTestVerifier` in #1689  For example, with the following,

```
        verifier.hasDependency("io.micronaut.oraclecloud", "micronaut-oraclecloud-function-http", Scope.COMPILE)
        !verifier.hasDependency("io.micronaut.oraclecloud", "micronaut-oraclecloud-function")
```

The second check fails because the check for "micronaut-oraclecloud-function" is matching "micronaut-oraclecloud-function-http". 

This fixes the regex in `GradleBuildTestVerifier.hasDependency()` to restore previous behavior while also supporting the change as intended?

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/build/dependencies/MicronautDependencyUtils.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/dependencies/MicronautDependencyUtils.java
@@ -22,6 +22,7 @@ public final class MicronautDependencyUtils {
     public static final String GROUP_ID_MICRONAUT_AWS = "io.micronaut.aws";
     public static final String GROUP_ID_MICRONAUT_CRAC = "io.micronaut.crac";
     public static final String GROUP_ID_MICRONAUT_GCP = "io.micronaut.gcp";
+    public static final String GROUP_ID_MICRONAUT_OCI = "io.micronaut.oraclecloud";
     public static final String GROUP_ID_MICRONAUT_SERDE = "io.micronaut.serde";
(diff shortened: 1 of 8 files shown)


alvarosanchez (approved):

sdelamo:
> Fixing missing dependencies for Gradle projects with "oracle-function-http" feature, to make consistent with Maven projects. 

It is not a bug in gradle. The [gradle plugin adds those dependencies automatically.](https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/3.7.x/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java).

sdelamo:
> This also fixes a bug from a change to `GradleBuildTestVerifier` in #1689 For example, with the following,
> 
> ```
>         verifier.hasDependency("io.micronaut.oraclecloud", "micronaut-oraclecloud-function-http", Scope.COMPILE)
>         !verifier.hasDependency("io.micronaut.oraclecloud", "micronaut-oraclecloud-function")
> ```
> 
> The second check fails because the check for "micronaut-oraclecloud-function" is matching "micronaut-oraclecloud-function-http".
> 
> This fixes the regex in `GradleBuildTestVerifier.hasDependency()` to restore previous behavior while also supporting the change as intended?

I added a test for the regex.

PR #1710, opened 2023-04-17
refactor: move GE Feature out of gradle package

timyates:
This used to be Gradle only.  For 3.9.0 it also works for Maven.

This PR moves it out of the gradle package into the parent, where it's more logically correct


sdelamo (changes requested):
Lets do this in Micronaut Starter 4.0.x. It is a breaking change.

PR #1717, opened 2023-04-17
Add feature class for Micronaut Validation

wetted:
This PR has two motivations:

1. remove static dependency notation from Gradle and Maven Rocker template. See #1707
2. prepare for Micronaut 4, where the dependency notation changes, and Micronaut Validation is in it's own module with changed dependencies, and documentation we can link in the readme for generated projects. See #1698

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/validator/HibernateValidationFeature.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/validator/HibernateValidationFeature.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 original authors
+ * Copyright 2017-2022 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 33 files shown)


wetted:
Moving this to draft for now. I think I'll have some more commits to add.

graemerocher:
does it need to be a default feature in 4.x?

sdelamo:
> does it need to be a default feature in 4.x?

I don't think so. I think we should keep the number of default features to a minimum

starter-core/src/main/java/io/micronaut/starter/feature/validator/HibernateValidationFeature.java
```diff
@@ -13,17 +13,16 @@
...
 
 import io.micronaut.starter.application.ApplicationType;
 import io.micronaut.starter.application.generator.GeneratorContext;
 import io.micronaut.starter.build.dependencies.Dependency;
 import io.micronaut.starter.feature.Category;
-import io.micronaut.starter.feature.Feature;
 import jakarta.inject.Singleton;
 
 @Singleton
-public class HibernateValidator implements Feature {
```
sdelamo: Please, don't rename the feature. We are trying to avoid unnecessary breaking changes if possible.

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/validator/HibernateValidationFeature.java
```diff
@@ -13,17 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micronaut.starter.feature.other;
+package io.micronaut.starter.feature.validator;
```
sdelamo: dont change the package.

starter-core/src/main/java/io/micronaut/starter/feature/validator/MicronautValidationFeature.java
```diff
@@ -0,0 +1,61 @@
...
+        return "Adds support for Micronaut Validation";
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.addDependency(MICRONAUT_VALIDATION_COMPILE);
+    }
+
+    @Override
+    public String getCategory() {
```
burtbeckwith: this isn't needed - it's the same value as in the interface

starter-core/src/main/java/io/micronaut/starter/feature/validator/HibernateValidator.java
```diff
@@ -13,17 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micronaut.starter.feature.other;
+package io.micronaut.starter.feature.validator;
```
sdelamo: dont change the package

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/validator/HibernateValidator.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-starter/pull/1717/commits/52687ebdd4cf8c507d807c2d7e4d74ff1124088d

starter-core/src/main/java/io/micronaut/starter/feature/validator/HibernateValidationFeature.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-starter/pull/1717/commits/52687ebdd4cf8c507d807c2d7e4d74ff1124088d

sdelamo (approved):

sdelamo (approved):

PR #1719, opened 2023-04-18
Refactor conditional 'http-client' test dependency from gradle/pom rocker templates into Java API.

wetted:
This is an attempt to refactor the 'http-client' test dependency from the pom/gradle Rocker files into the starter Java API. This one in particular is a conditional dependency that doesn't apply specifically to any applied features. The challenge is that a dependency like this has to be applied after all features have been added and 'applied' (`generatorContext.applyFeatures()`) but before the rocker templates are rendered.

In this PR I do just that in `DefaultProjectGenerator.generate()`. That works well enough for Gradle projects (the dependency is added correctly), but doesn't for Maven, where it seems Maven dependencies have already been resolved, and the dependency is never added to the build. I also tried adding the dependencies in the `apply` methods of the `Maven` and `Gradle` `BuildFeature` classes, but that is too early, since not all features have been resolved/applied, and the dependency ends up included where it should not be.

What I have here is a bit of a smell. I tried some other things without success. Any ideas @sdelamo


sdelamo:
closed via https://github.com/micronaut-projects/micronaut-starter/pull/1725

PR #1728, opened 2023-04-24
Properties default Configuration Feature

sdelamo:


timyates (approved):
I switched to using Yaml.NAME in the tests and added Toml.NAME and Properties.NAME

Also fixed ConfigurationSpec and SwaggerUISpec in test-features

I skipped CI, as I didn't want to use all the starter agents incase you were doing something @sdelamo

PR #1729, opened 2023-04-24
disable Jansi

sdelamo:


graemerocher (approved):

starter-core/src/main/java/io/micronaut/starter/feature/logging/Logback.java
```diff
@@ -76,7 +77,10 @@ protected boolean useJansi(@NonNull GeneratorContext generatorContext) {
         if (generatorContext.getOperatingSystem() == OperatingSystem.WINDOWS) {
             return false;
         }
-        return !generatorContext.getFeatures().hasFeature(AwsLambda.class);
+        if (generatorContext.getFeatures().hasFeature(AwsLambda.class)) {
+            return false;
+        }
+        return USE_JANSI;
```
timyates, line 83: So, always `false`?  Shouldn't we just replace the whole method?

Or is this temporary?

PR #1730, opened 2023-04-24
Hide community libraries

sdelamo:
They will need to be recompile for Micronaut Framework 4


timyates (approved):
👍 

I simplified things by only adding the `@Requires` at spec level instead of a mix of that and the feature level

nbrugger-tgm:
If 3rd party libraries need to recompile anyway would https://github.com/micronaut-projects/micronaut-core/issues/6395#issuecomment-1518619408 be on the table again?

Sorry for jumping in here. I am not inpatient, I can wait - but (the linked issue) is a breaking change and therefore won't happen after 4.0 release

PR #1731, opened 2023-04-24
remove @Ignore GCP

sdelamo:


starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildGradle.rocker.raw
```diff
@@ -74,16 +74,24 @@ java {
...
 @if (features.language().isKotlin()) {
     compileKotlin {
-        kotlinOptions {
-            jvmTarget = "@features.getTargetJdk()"
+        compilerOptions {
+@if(gradleBuild.getDsl() == GradleDsl.KOTLIN) {
+            jvmTarget.set(org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_@features.getTargetJdk())
+} else {
+           jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_@features.getTargetJdk()
+}
```
timyates, line 82: This only [goes up to 19](https://github.com/JetBrains/kotlin/blob/1.8.0/libraries/tools/kotlin-gradle-compiler-types/src/generated/kotlin/org/jetbrains/kotlin/gradle/dsl/JvmTarget.kt) at time of writing...

We should be ok, as we [only go up to 17](https://github.com/micronaut-projects/micronaut-starter/blob/9c65de995e6035a1bc3e2b4a74018551a30ec3b9/starter-core/src/main/java/io/micronaut/starter/options/JdkVersion.java#L29-L31)

Actually I guess for 4.0.x, this should be changed to just include 17?

timyates (approved):

PR #1739, opened 2023-04-24
Add feature for HTTP Client JDK

### LINKED ISSUE #1726: Add feature to add HTTP Client JDK
https://docs.micronaut.io/snapshot/guide/#jdkHttpClient

PR should be done against 4.0.x

wetted:
closes #1726

@sdelamo @timyates 
I am not entirely sure about two things for this PR:
1. whether the behavior of the Netty 'http-client' feature as a default feature should change (I didn't change it)
2. what application types this feature supports (I made it all but FUNCTION).

### DIFF
--- a/starter-core/src/test/groovy/io/micronaut/starter/feature/httpclient/HttpClientFeatureSpec.groovy
+++ b/starter-core/src/test/groovy/io/micronaut/starter/feature/httpclient/HttpClientFeatureSpec.groovy
@@ -1,0 +1,16 @@
+package io.micronaut.starter.feature.httpclient
+
+import io.micronaut.starter.ApplicationContextSpec
+import io.micronaut.starter.fixture.CommandOutputFixture
+
+class HttpClientFeatureSpec extends ApplicationContextSpec implements CommandOutputFixture {
+
+    void 'test only one of the HttpClient features can be selected'() {
+        when:
+        generate(['http-client', 'http-client-jdk'])
+
+        then:
+        def e = thrown(IllegalArgumentException)
+        e.message.startsWith("There can only be one of the following features selected:")
+    }
+}
--- a/starter-core/src/test/groovy/io/micronaut/starter/feature/rxjava/RxJava3Spec.groovy
+++ b/starter-core/src/test/groovy/io/micronaut/starter/feature/rxjava/RxJava3Spec.groovy
@@ -3,6 +3,9 @@ package io.micronaut.starter.feature.rxjava
 import io.micronaut.starter.ApplicationContextSpec
 import io.micronaut.starter.BuildBuilder
 import io.micronaut.starter.application.ApplicationType
+import io.micronaut.starter.build.BuildTestUtil
+import io.micronaut.starter.build.BuildTestVerifier
+import io.micronaut.starter.build.dependencies.Scope
 import io.micronaut.starter.feature.Category
 import io.micronaut.starter.fixture.CommandOutputFixture
 import io.micronaut.starter.options.BuildTool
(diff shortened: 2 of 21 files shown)


sdelamo:
@timyates can you review?

closes https://github.com/micronaut-projects/micronaut-starter/issues/1726


> whether the behavior of the Netty 'http-client' feature as a default feature should change (I didn't change it)

I have changed it so that by default adds the client to the test classpath. 

> what application types this feature supports (I made it all but FUNCTION).

Everyone

starter-core/src/test/groovy/io/micronaut/starter/feature/rxjava/RxJava3Spec.groovy
```diff
@@ -42,46 +45,19 @@ class RxJava3Spec extends ApplicationContextSpec implements CommandOutputFixture
...
 
     @Unroll
-    void 'dependency is included with maven and feature rxjava3 for language=#language'(Language language) {
+    void 'dependency is included with gradle and feature rxjava3 for language=#language'(Language language, BuildTool buildTool) {
         when:
-        String template = new BuildBuilder(beanContext, BuildTool.MAVEN)
-                .features(['rxjava3'])
+        String template = new BuildBuilder(beanContext, buildTool)
                 .language(language)
+                .features(['http-client', 'rxjava3'])
```
timyates, line 52: If we select `['http-client-jdk', 'rxjava3']` we end up with

```
    implementation("io.micronaut:micronaut-http-client-jdk")
    implementation("io.micronaut.rxjava3:micronaut-rxjava3")
```

Which I believe is wrong...  We should probably not allow this combination?

starter-core/src/test/groovy/io/micronaut/starter/feature/rxjava/RxJava3Spec.groovy
```diff
@@ -42,46 +45,19 @@ class RxJava3Spec extends ApplicationContextSpec implements CommandOutputFixture
     }
 
     @Unroll
-    void 'dependency is included with maven and feature rxjava3 for language=#language'(Language language) {
+    void 'dependency is included with gradle and feature rxjava3 for language=#language'(Language language, BuildTool buildTool) {
```
timyates:
```suggestion
    void 'dependency is included with #buildTool and feature rxjava3 for language=#language'(Language language, BuildTool buildTool) {
```

starter-core/src/test/groovy/io/micronaut/starter/feature/httpclient/HttpClientFeatureSpec.groovy
```diff
@@ -0,0 +1,16 @@
...
+
+class HttpClientFeatureSpec extends ApplicationContextSpec implements CommandOutputFixture {
+
+    void 'test only one of the HttpClient features can be selected'() {
+        when:
+        generate(['http-client', 'http-client-jdk'])
+
+        then:
+        def e = thrown(IllegalArgumentException)
+        e.message.startsWith("There can only be one of the following features selected:")
```
timyates, line 14: Why do we have this restriction?  I thought we added tests to make sure it worked with both on the classpath?

timyates (changes requested):

starter-core/src/test/groovy/io/micronaut/starter/feature/rxjava/RxJava3Spec.groovy
timyates: Or maybe we should...  So many options...

timyates (approved):

PR #1740, opened 2023-04-25
Groovy for custom aws runtime

sdelamo:


sdelamo:
With the release of AWS M2, these tests now pass.

PR #1741, opened 2023-04-25
Bug fix for recreation of new the file in case that the file already …

nenadvico:
Bug fix for recreation of new the file in case that the file already exists
It prevents the issue that content of previous file appeared at the end of the new file (in case that new content is shorter then previous one)


sdelamo:
> Bug fix for recreation of new the file in case that the file already exists It prevents the issue that content of previous file appeared at the end of the new file (in case that new content is shorter then previous one)

do you have a test or an scenario to reproduce the issue?

nenadvico:
We are using it in Oracle internal projects and it cause one issue.
I will try to make the test to reproduce it and cover that case

nenadvico:
I add test which could reproduce the case

sdelamo (approved):

PR #1753, opened 2023-04-28
support Java 17 runtime in CDK

sdelamo:
https://github.com/micronaut-projects/micronaut-gradle-plugin/actions?query=branch%3A3.7.x


timyates (approved):
Ran a build locally via the CLI and Java 17 was set as the default and configured in the CDK 👍 

Couldn't run the CDK as I don't have the CDK set up locally 🫤

PR #1757, opened 2023-05-01
fix: for SQLServer generate test-resources accept license

sdelamo:
Currently, we are doing the generation of the property directly in the tests. I don't know why we did it that way. The generated applications should include this property.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/database/DatabaseDriverConfigurationFeature.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/database/DatabaseDriverConfigurationFeature.java
@@ -45,7 +45,7 @@ public interface DatabaseDriverConfigurationFeature extends Feature {
         }
         Optional.ofNullable(dbFeature.getDriverClass()).ifPresent(driver -> config.put(getDriverKey(), driver));
         dbFeature.getDbType().ifPresent(dbType -> config.put(PROPERTY_DATASOURCES_DEFAULT_DB_TYPE, dbType.toString()));
-        final Map<String, Object> additionalConfig = dbFeature.getAdditionalConfig();
+        final Map<String, Object> additionalConfig = dbFeature.getAdditionalConfig(generatorContext);
         if (!additionalConfig.isEmpty()) {
             config.putAll(additionalConfig);
         }
--- a/starter-core/src/main/java/io/micronaut/starter/feature/database/DatabaseDriverFeature.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/database/DatabaseDriverFeature.java
@@ -104,7 +104,7 @@ public abstract class DatabaseDriverFeature extends EaseTestingFeature implement
         return Optional.empty();
     }
 
-    public Map<String, Object> getAdditionalConfig() {
+    public Map<String, Object> getAdditionalConfig(GeneratorContext generatorContext) {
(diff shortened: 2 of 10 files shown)


sdelamo:
it seems it failed with 

```


Message: Could not resolve placeholder ${auto.test.resources.jpa.default.properties.hibernate.connection.url} |  
-- | --
Path Taken: SessionFactory.buildHibernateSessionFactoryBuilder(SessionFactoryBuilder sessionFactoryBuilder) --> SessionFactory.buildHibernateSessionFactoryBuilder([SessionFactoryBuilder sessionFactoryBuilder]) --> SessionFactoryBuilder.buildHibernateSessionFactoryBuilder(Metadata metadata,[JpaConfiguration jpaConfiguration]) |  
Caused by: io.micronaut.context.exceptions.BeanInstantiationException: |  
Error instantiating bean of type  [org.hibernate.boot.SessionFactoryBuilder] |  
  |  
Message: Could not resolve placeholder ${auto.test.resources.jpa.default.properties.hibernate.connection.url} |  
Path Taken: SessionFactory.buildHibernateSessionFactoryBuilder(SessionFactoryBuilder sessionFactoryBuilder) --> SessionFactory.buildHibernateSessionFactoryBuilder([SessionFactoryBuilder sessionFactoryBuilder]) --> SessionFactoryBuilder.buildHibernateSessionFactoryBuilder(Metadata metadata,[JpaConfiguration jpaConfiguration]) |  
Caused by: io.micronaut.context.exceptions.ConfigurationException: Could not resolve placeholder ${auto.test.resources.jpa.default.properties.hibernate.connection.url} |  
```

melix:
The error is : https://ge.micronaut.io/s/agt3lan3tfiyy/tests/:test-features:test/io.micronaut.starter.core.test.feature.database.HibernateReactiveJpaSpec/test%20maven%20hibernate-reactive-jpa%20with%20java%20and%20sqlserver?focused-exception-line=2-54&top-execution=3

> You must set the property 'test-resources.containers.mssql.accept-license' to true in order to use a Microsoft SQL Server test container

That said I disagree that the user apps should have this property set: it's not our responsibility to accept a license on behalf of the user, we would be at legal risk. At best, we should have `test-resources.containers.mssql.accept-license=false` to the generated template, so that the user only has to switch from `false` to `true.

sdelamo:
@wetted can you change this PR so that as suggested by @melix  we generate it by default to `false`. 
The, create an invisible feature, place it in `test-features`. The hidden feature should replaces SQLServer and override a method  accept the license:

```
@Replaces(SQLServer.class)
@Singleton
public class AcceptLicenseSQLServer extends SQLServer {
@Override
protected boolean acceptLicense() {
return true;
}
...
..
.
}
```

wetted:
@sdelamo I have the replacement feature working, but there's another issue where the test-resources property isn't generated at all when `HibernateReactive` features are included, even though the `test-resources` feature is...investigating.

wetted:
I fixed the problem with making sure the "test-resources.containers.mssql.accept-license" is applied correctly, including for when `DataHibernateReactive` and `HibernateReactiveJpa`. Previously it was only applied at all if a `DatabaseDriverConfigurationFeature` was present.

Now the tests have different issues. Investigating.

wetted:
ping @sdelamo this is good to go now.

PR #1758, opened 2023-05-01
Maven IncompatibleClassChangeError: class io.micronaut.http.filter.GenericHttpFilter$TerminalWithReactorContext cannot extend sealed interface io.micronaut.http.filter.GenericHttpFilter

sdelamo:


sdelamo:
Executing the tests is slow. The easiest way to reproduce is to generate an app with the same features via: 

```
./gradlew micronaut-cli:run --args="create-app --jdk 17 -b maven -f \"XXX\" temp
```

alvarosanchez:
This is fixed in Maven Plugin 4.0.0-M3, which will arrive once https://github.com/micronaut-projects/micronaut-platform/pull/301 is merged

PR #1764, opened 2023-05-01
remove @ignore

sdelamo:


sdelamo:
@timyates check this out. I have not gone into a lot of detail. But it seems it does not stop in the Exception Handler.

timyates:
@sdelamo There's no body in the error response...

The exception has the message `Client '/diff': The requested feature does not exist: junkkkkk`, but we're not returning a json error body 🤔

timyates:
We seem to have a `ErrorResponseProcessor` (an instance of HateoasErrorResponseProcessor) in the bean context of the app under test, but it's not being called 🤔

timyates:
I currently don't know why the HttpStatusHandler is not invoked

graemerocher:
@yawkat is there some regression here with exception handlers?

timyates:
@yawkat ping 😉 https://github.com/micronaut-projects/micronaut-starter/pull/1764#issuecomment-1532627477

yawkat:
it's a core bug

sdelamo:
This should be fix with 4.0.0-M6 core.

sdelamo:
this still fails

yawkat:
everything fails now, it's another classpath issue in the upgrade

timyates:
@yawkat It's using M6 properly now after the merge, but it still fails 🤔

yawkat:
the patch is not included in M6

timyates:
Ahhhhhhh.... 😊

timyates:
@yawkat Ok still fails [with M7](https://ge.micronaut.io/s/7q5nwbeehfic4/dependencies?focusedDependency=WzEsNiw4NTIsWzEsNixbODUyXV1d&toggled=W1sxXSxbMSw2XV0)

[This test still fails](https://github.com/micronaut-projects/micronaut-starter/blob/7219220384fee9873abbd8e28fa5b886bc6acb9b/starter-api/src/test/groovy/io/micronaut/starter/api/DiffControllerSpec.groovy#L60) as it's expecting a Hateoas error response

But it gets a plain string `The requested feature does not exist: junkkkkk`

https://ge.micronaut.io/s/7q5nwbeehfic4/tests/:starter-api:test/io.micronaut.starter.api.DiffControllerSpec/test%20diff%20invalid%20feature?top-execution=1

timyates:
Discussed with @yawkat and the issue seems to be that for some reason with 4.0.x we now end up with a JustThrowPublisher instead of just throwing the exception.

The publisher handler does not handle JsonError encoding via the error processors.

Jonas suspects that something in AOP has changed this, and we should switch the test to just checking for the String response.

I've had a look at the [starter-ui error handling](https://github.com/micronaut-projects/micronaut-starter-ui/blob/16c98494650a71115d3f5892a2acc93898519443/app/launch/src/components/ErrorView/ErrorViewData.js#L76), and I *think* this would still work, however I'm going to try to get a reproducer together so I can raise an issue in core against the 4.0 Project so we don't lose track of this...

timyates:
I have created a reproducer and raised an issue in core to look into this https://github.com/micronaut-projects/micronaut-core/issues/9418

We could fix it here by building a String instead of using diffFlowable to create a Publisher<String>, but it's quite good that this was caught by us using a Publisher

sdelamo:
Thanks @timyates and @yawkat. I think this just passes for RC1 core

PR #1838, opened 2023-05-26
Move CDK module to starter

sdelamo:
We deleted from aws here https://github.com/micronaut-projects/micronaut-aws/pull/1650

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/aws/Cdk.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/aws/Cdk.java
@@ -79,6 +79,20 @@ public class Cdk implements MultiProjectFeature, InfrastructureAsCodeFeature {
     private static final String MAIN_CLASS_NAME = "Main";
     private static final String ARTIFACT_ID_AWS_CDK_LIB = "aws-cdk-lib";
     private static final String GROUP_ID_SOFTWARE_AMAZON_AWSCDK = "software.amazon.awscdk";
+    private static final String ARTIFACT_ID_MICRONAUT_STARTER_AWS_CDK = "micronaut-starter-aws-cdk";
+    private static final Dependency DEPENDENCY_CDK =
+            MicronautDependencyUtils.awsDependency()
+                    .artifactId("micronaut-aws-cdk")
+                    .version("4.0.0-M2")
+//            MicronautDependencyUtils.starterDependency()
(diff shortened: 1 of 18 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/aws/Cdk.java
```diff
@@ -79,6 +79,20 @@ public class Cdk implements MultiProjectFeature, InfrastructureAsCodeFeature {
     private static final String MAIN_CLASS_NAME = "Main";
     private static final String ARTIFACT_ID_AWS_CDK_LIB = "aws-cdk-lib";
     private static final String GROUP_ID_SOFTWARE_AMAZON_AWSCDK = "software.amazon.awscdk";
+    private static final String ARTIFACT_ID_MICRONAUT_STARTER_AWS_CDK = "micronaut-starter-aws-cdk";
+    private static final Dependency DEPENDENCY_CDK =
+            MicronautDependencyUtils.awsDependency()
+                    .artifactId("micronaut-aws-cdk")
+                    .version("4.0.0-M2")
+//            MicronautDependencyUtils.starterDependency()
```
sdelamo, line 87: For M4, we need to delete 84-87 and uncomment the following lines

timyates:
I think these have been failing in starter all week

https://ge.micronaut.io/s/ntn25ksf3tbgc/tests/overview?outcome=failed

They're the failures in the CDK PR https://github.com/micronaut-projects/micronaut-starter/pull/1838#pullrequestreview-1446305015

But they happen on 4.0.x of starter as well

The `io.micronaut.starter.api.create.github.GitHubCreateControllerSpec` passes with M3, but not M5

The others I can't get to pass with M3

The Azure one is failing with

```
Caused by: java.lang.IncompatibleClassChangeError: class io.micronaut.servlet.http.ServletBodyBinder overrides final method io.micronaut.http.bind.binders.DefaultBodyAnnotationBinder.getAnnotationType()Ljava/lang/Class;
```

I think it's because it's pulling in M2 Azure which is based on an old version of Servlet

It's fixed by doing

```
configurations.all {
    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
        if (details.requested.name == 'micronaut-azure-function-http') {
            details.useVersion "5.0.0-M3"
            details.because "We need a newer version"
        }
    }
}
```

to get the latest Milestone

But the Github one feels like a bug in core

timyates (approved):

PR #1852, opened 2023-06-06
Starter GithubCreateControllerSpec fails

sdelamo:


timyates:
So all of these tests are failing with "Failed to decode the body for the given content type [application/vnd.github.v3+json]"

🤔

timyates:
Cancelled the workflows as I don't want to block any release that's going on

timyates:
@sdelamo This now passes locally

PR #1853, opened 2023-06-06
Starter FunctionSpec fails

sdelamo:


timyates:
Ahhhh...  This is because we _used_ to add "application/zip" to the base64 content types

https://github.com/micronaut-projects/micronaut-aws/blob/07c4c6aab42e1a68adf98c73fcda736fa6a4822a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/AbstractLambdaContainerHandler.java#L110

But now we just have a static list of content types that doesn't include "application/zip"

timyates:
This should fix the test (but will require another milestone of aws) https://github.com/micronaut-projects/micronaut-aws/pull/1754

timyates:
Just tested with

```
configurations.all {
    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
        if (details.requested.group == 'io.micronaut.aws') {
            details.useVersion('4.0.0-M7')
        }
    }
}
```

And this passes again 🎉

PR #1865, opened 2023-06-14
Change handler code

sdelamo:


starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/AnalyticsController.java
```diff
@@ -25,12 +25,12 @@
...
 import io.micronaut.starter.analytics.Generated;
 
-import javax.transaction.Transactional;
+import jakarta.transaction.Transactional;
 import java.util.List;
 import java.util.stream.Collectors;
 
 @Controller("/analytics")
-@ExecuteOn(TaskExecutors.IO)
+@ExecuteOn(TaskExecutors.BLOCKING)
```
timyates, line 33: Was this wrong before?

timyates (approved):

starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/AnalyticsController.java
sdelamo: No. But we recommend people to use `BLOCKING` going forward.

[Virtual Threads](https://docs.micronaut.io/4.0.0-M7/guide/#virtualThreads)

> The Micronaut framework will detect virtual thread support and use it for the executor named blocking if available. If virtual threads are not supported, this executor will be aliased to the io thread pool.

PR #1913, opened 2023-06-26
move enforcer as a default feature

### LINKED ISSUE #1712: Update Starter to reflect the changes in Micronaut Maven Plugin 4.x
- New group id: `io.micronaut.maven`.
- Should add `org.apache.maven.plugins:maven-enforcer-plugin` by default

sdelamo:
Close #1712


alvarosanchez:
Does this mean that the user might be able to de-select the feature? I don't think this should be visible at all, same as they can't toggle the Micronaut Maven plugin

sdelamo:
> Does this mean that the user might be able to de-select the feature? I don't think this should be visible at all, same as they can't toggle the Micronaut Maven plugin

I made it not visible.

alvarosanchez (approved):

PR #1923, opened 2023-06-29
add object-storage-local feature

sdelamo:


alvarosanchez (commented):
Good catch, thank you!

Should we add this feature in the test scope by default?

alvarosanchez:
@sdelamo I think we should add this feature in either test or development scopes by default

PR #1917, opened 2023-06-27
Fix generated Azure projects in starter [skip ci]

timyates:
For Azure 5.0.0 things have moved, and we need to change the generated test

This PR also removes the instructions to `clean` all the time with Gradle in the README

It will not pass until https://github.com/micronaut-projects/micronaut-azure/pull/509 is merged and released to platform.

PR #1925, opened 2023-06-29
Remove java8/java11 requirement in tests

timyates:
I think a lot of this is due to the cloud providers (at the time) not having java 17 support. This is no longer the case, so we should be able to remove the java8/java11 requirement so that the tests run again.

Queued up after https://github.com/micronaut-projects/micronaut-starter/pull/1917 as there are azure tests that will fail without it

### DIFF
--- a/starter-core/src/test/groovy/io/micronaut/starter/feature/graalvm/GraalVMSpec.groovy
+++ b/starter-core/src/test/groovy/io/micronaut/starter/feature/graalvm/GraalVMSpec.groovy
@@ -10,7 +10,6 @@ import spock.lang.Shared
 import spock.lang.Subject
 import spock.lang.Unroll
 
-@Requires({ jvm.isJava8() || jvm.isJava11() })
 class GraalVMSpec extends ApplicationContextSpec implements CommandOutputFixture {
 
     @Subject
@@ -63,7 +62,7 @@ class GraalVMSpec extends ApplicationContextSpec implements CommandOutputFixture
             <path>
               <groupId>io.micronaut</groupId>
               <artifactId>micronaut-graal</artifactId>
-              <version>\${micronaut.version}</version>
+              <version>\${micronaut.core.version}</version>
             </path>
 """)
         template.contains("""
@@ -91,7 +90,7 @@ class GraalVMSpec extends ApplicationContextSpec implements CommandOutputFixture
                <annotationProcessorPath>
                  <groupId>io.micronaut</groupId>
                  <artifactId>micronaut-graal</artifactId>
-                 <version>${micronaut.version}</version>
+                 <version>${micronaut.core.version}</version>
                </annotationProcessorPath>
 ''')
         template.contains("""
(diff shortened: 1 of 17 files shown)


test-buildtool/src/test/groovy/io/micronaut/starter/core/test/buildTool/MavenPackageSpec.groovy
```diff
@@ -63,15 +60,15 @@ class MavenPackageSpec extends CommandSpec {
...
+        output.contains("Using BASE_IMAGE: ghcr.io/graalvm/native-image:ol8-java17-22.3.0")
         
         where:
         lang << Language.values()
     }
 
     @IgnoreIf({ BuildToolTest.IGNORE_MAVEN })
     @Unroll
-    @Requires({ Jvm.current.java8 || Jvm.current.java11 })
+    @PendingFeature
```
timyates: @alvarosanchez I had to mark this as pending, as it doesn't pass...

Is this a valid test anymore?

starter-core/src/test/groovy/io/micronaut/starter/feature/graalvm/GraalVMSpec.groovy
```diff
@@ -91,7 +90,7 @@ class GraalVMSpec extends ApplicationContextSpec implements CommandOutputFixture
                <annotationProcessorPath>
                  <groupId>io.micronaut</groupId>
                  <artifactId>micronaut-graal</artifactId>
-                 <version>${micronaut.version}</version>
+                 <version>${micronaut.core.version}</version>
```
timyates, line 92: @alvarosanchez Is this right?  I think so...  at least, that's what we generate

sdelamo (approved):

sdelamo:
@alvarosanchez ping! see comments above from @timyates 👆🏻

starter-core/src/test/groovy/io/micronaut/starter/feature/graalvm/GraalVMSpec.groovy
alvarosanchez: Correct

alvarosanchez:
@timyates @sdelamo I fixed the tests

alvarosanchez (approved):

timyates:
@alvarosanchez Some maven failures...

https://ge.micronaut.io/s/6p7zgh6fbougc/tests/overview?outcome=failed

Seems to be `'gu' tool wasn't found. This probably means that JDK at isn't a GraalVM distribution.`

Any ideas?

alvarosanchez:
There are 2 different things:

- Docker Native packaging fails because it doesn't finish before the 180s default timeout. Perhaps it's a matter of changing the `BUILD SUCESS` assertion for a different text that shows up earlier, and assume it'll finish ok.
- Native Image packaging fails because we're using Eclipse Temurin 😅 . So either change the JDK or add a requirement

timyates:
@alvarosanchez That should do it 🤞

alvarosanchez (approved):

PR #1956, opened 2023-07-10
AOT as a Default feature

### LINKED ISSUE #1804: Include Micronaut AOT by default in 4.x
Micronaut AOT should be promoted to a default feature.

sdelamo:
Close https://github.com/micronaut-projects/micronaut-starter/issues/1804


starter-core/src/test/groovy/io/micronaut/starter/feature/messaging/kafka/KafkaSpec.groovy
```diff
@@ -62,7 +62,7 @@ class KafkaSpec extends ApplicationContextSpec implements CommandOutputFixture {
         template.contains('''<artifactId>micronaut-maven-plugin</artifactId>
           <configuration>
             <shared>true</shared>
-          </configuration>''')
```
sdelamo, line 65: I had to remove this because config with AOT enabled looks like: 

```xml
    <artifactId>micronaut-maven-plugin</artifactId>
          <configuration>
            <shared>true</shared>
            <configFile>aot-${packaging}.properties</configFile>
          </configuration>
```

starter-core/src/test/groovy/io/micronaut/starter/feature/messaging/kafka/KafkaStreamsSpec.groovy
```diff
@@ -83,7 +83,7 @@ class KafkaStreamsSpec extends ApplicationContextSpec implements CommandOutputFi
         template.contains('''<artifactId>micronaut-maven-plugin</artifactId>
           <configuration>
             <shared>true</shared>
-          </configuration>''')
```
sdelamo, line 86: I had to remove this because config with AOT enabled looks like: 

```xml
    <artifactId>micronaut-maven-plugin</artifactId>
          <configuration>
            <shared>true</shared>
            <configFile>aot-${packaging}.properties</configFile>
          </configuration>
```

starter-core/src/test/groovy/io/micronaut/starter/feature/messaging/mqtt/MqttSpec.groovy
```diff
@@ -60,7 +60,7 @@ class MqttSpec extends ApplicationContextSpec implements CommandOutputFixture {
         template.contains('''<artifactId>micronaut-maven-plugin</artifactId>
           <configuration>
             <shared>true</shared>
-          </configuration>''') == (dependency == 'mqttv5')
```
sdelamo: I had to remove this because config with AOT enabled looks like: 

```xml
    <artifactId>micronaut-maven-plugin</artifactId>
          <configuration>
            <shared>true</shared>
            <configFile>aot-${packaging}.properties</configFile>
          </configuration>
```

starter-core/src/test/groovy/io/micronaut/starter/feature/messaging/rabbitmq/RabbitMQSpec.groovy
```diff
@@ -46,7 +46,7 @@ class RabbitMQSpec extends ApplicationContextSpec implements CommandOutputFixtur
         template.contains('''<artifactId>micronaut-maven-plugin</artifactId>
           <configuration>
             <shared>true</shared>
-          </configuration>''')
```
sdelamo, line 49: I had to remove this because config with AOT enabled looks like: 

```xml
    <artifactId>micronaut-maven-plugin</artifactId>
          <configuration>
            <shared>true</shared>
            <configFile>aot-${packaging}.properties</configFile>
          </configuration>
```

melix (approved):
I'm still wondering if we should have the default values to `true`. Of course it would mean that AOT would do nothing by default, so we'd have AOT enabled by default but doing nothing, but it would be safer. Or we wait for bug reports :)

sdelamo:
> I'm still wondering if we should have the default values to `true`. Of course it would mean that AOT would do nothing by default, so we'd have AOT enabled by default but doing nothing, but it would be safer. Or we wait for bug reports :)

I don't see a lot of risk because AOT is opt-in. Is not it?. If I run `./gradlew build`, two fat jars get generated `all.jar` and `optimized-all.jar`. So the user has to optin to use the optimized fat Jar. The same applies for native image generation and docker images generation. AOT does not replace existing gradle tasks a developer may already use but adds extra tasks.

For Maven,  the user has to optin in with `-Dmicronaut.aot.enabled=true`.

melix:
It's enabled but it's true that you have to consume it explicitly (e.g run `optimizedJitJarAll`)

alvarosanchez:
@sdelamo `micronaut.aot.enabled` defaults to true: https://github.com/micronaut-projects/micronaut-starter/blob/4.0.x/starter-core/src/main/java/io/micronaut/starter/feature/build/MicronautAot.java#L94

So isn't the point of #1804 to produce optimized applications by default? If so, then AOT should be default **and enabled**

melix:
> So isn't the point of https://github.com/micronaut-projects/micronaut-starter/issues/1804 to produce optimized applications by default?

I think the logic is quite different for Maven and Gradle. In Gradle it's additional binaries that are produced, that you _can_ use. So it's quite harmless to enable by default. For Maven this would replace what you normally produce, so the risks of breaking because you simply don't understand what AOT is is much higher.

alvarosanchez:
Then perhaps the correct solution would be to change Maven plugin to generate both artifacts. I'm not sure now if this is at all possible, so if it is not, then at least we should set `micronaut.aot.enabled` to false by default, and let them opt in

sdelamo:
> Then perhaps the correct solution would be to change Maven plugin to generate both artifacts. I'm not sure now if this is at all possible, so if it is not, then at least we should set `micronaut.aot.enabled` to false by default, and let them opt in

[I have changed Maven to generate aot=false by default](https://github.com/micronaut-projects/micronaut-starter/pull/1956/commits/39b1714d14056f5aa73b7b573fbfa58f40eb68b4). The maven test checks it is disabled if Aot feature is a default feature.

alvarosanchez (approved):

PR #1959, opened 2023-07-12
Micrometer features need r2dbc-pool runtime dependency when combined with r2dbc feature

### LINKED ISSUE #1535: r2dbc-pool not available to be selected at project creation, but wanted by micrometer
### Description

Launch does not provide an option to select R2DBC-Pool to be included in a project. 

Also, if the users selects r2dbc and also metrics, the generated application does not run, as `micronaut-micrometer-core`
expects `io.r2dbc.pool.ConnectionPool` to be available in runtime . 

### Task List

- [X] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [ ] Example that reproduces the problem uploaded to Github
- [X ] Full description of the issue provided (see below)

### Steps to Reproduce

1. On launch, select `data-r2dbc`, `postgres` and `micrometer-prometheus`
2. try to run the generated application

### Expected Behaviour

Application should start without errors

### Actual Behaviour

Application crashes with runtime error

### Environment Information

- **Operating System**: MacOS
- **Micronaut Version:** 3.7.3
- **JDK Version:** coreto-11

wetted:
fixes #1535

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/micrometer/Core.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/micrometer/Core.java
@@ -37,7 +37,6 @@ public class Core implements Feature {
             .runtime()
             .build();
 
-
     @Override
     public String getName() {
         return "micrometer";
--- a/starter-core/src/main/java/io/micronaut/starter/feature/micrometer/Core.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/micrometer/Core.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2022 original authors
+ * Copyright 2017-2023 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/micrometer/Core.java
```diff
@@ -20,10 +20,22 @@
 import io.micronaut.starter.build.dependencies.Dependency;
 import io.micronaut.starter.feature.Feature;
 
+import io.micronaut.starter.feature.database.r2dbc.R2dbcFeature;
 import jakarta.inject.Singleton;
 
 @Singleton
 public class Core implements Feature {
+    private static final Dependency MICRONAUT_MICROMETER_CORE = Dependency.builder()
+            .groupId("io.micronaut.micrometer")
```
sdelamo, line 29: This should use Micronaut Dependency Utils

sdelamo (approved):

PR #1967, opened 2023-07-17
Upgrade to Gradle plugin 4.0.1

melix:


sdelamo (changes requested):
Please, update it in every occurrence of the plugin https://github.com/micronaut-projects/micronaut-starter/blob/4.0.x/settings.gradle#L8

PR #1979, opened 2023-07-24
Add endpoint to download analytics as excel

### LINKED ISSUE #1976: Add endpoint to download analytics as excel
Create an endpoint in `starter-analytics-postgres` with route GET `/analytics/excel`. 

It should generate a spread sheet using [Apache POI](https://poi.apache.org/). 

It should generate a single sheet. 

it should have a header row

```
id | features| type | language | buildTool | testFramework | jdkVersion | micronautVersion | dateCreated
```

`features` should be a comma separated names of application features’ names.

guillermocalvo:
Closes #1976

### DIFF
--- a/starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/AnalyticsController.java
+++ b/starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/AnalyticsController.java
@@ -20,7 +20,6 @@ import io.micronaut.data.model.Pageable;
 import io.micronaut.http.HttpResponse;
 import io.micronaut.http.HttpStatus;
 import io.micronaut.http.MediaType;
-import io.micronaut.http.MutableHttpResponse;
 import io.micronaut.http.annotation.Body;
 import io.micronaut.http.annotation.Controller;
 import io.micronaut.http.annotation.Get;
--- a/starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/AnalyticsController.java
+++ b/starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/AnalyticsController.java
@@ -16,16 +16,25 @@
 package io.micronaut.starter.analytics.postgres;
 
 import io.micronaut.core.annotation.NonNull;
+import io.micronaut.data.model.Pageable;
+import io.micronaut.http.HttpResponse;
 import io.micronaut.http.HttpStatus;
+import io.micronaut.http.MediaType;
+import io.micronaut.http.MutableHttpResponse;
 import io.micronaut.http.annotation.Body;
 import io.micronaut.http.annotation.Controller;
 import io.micronaut.http.annotation.Get;
 import io.micronaut.http.annotation.Post;
+import io.micronaut.http.annotation.Produces;
+import io.micronaut.http.server.types.files.StreamedFile;
 import io.micronaut.scheduling.TaskExecutors;
 import io.micronaut.scheduling.annotation.ExecuteOn;
 import io.micronaut.starter.analytics.Generated;
 
 import jakarta.transaction.Transactional;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -35,12 +44,15 @@ public class AnalyticsController {
 
     private final ApplicationRepository applicationRepository;
     private final FeatureRepository featureRepository;
+    private final ExcelGenerator excelGenerator;
 
     public AnalyticsController(
             ApplicationRepository applicationRepository,
-            FeatureRepository featureRepository) {
+            FeatureRepository featureRepository,
+            ExcelGenerator excelGenerator) {
         this.applicationRepository = applicationRepository;
         this.featureRepository = featureRepository;
+        this.excelGenerator = excelGenerator;
     }
 
     @Get("/top/features")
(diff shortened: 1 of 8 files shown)


starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/AnalyticsController.java
```diff
@@ -93,4 +105,21 @@ HttpStatus applicationGenerated(@NonNull @Body Generated generated) {
...
         return HttpStatus.ACCEPTED;
     }
+
+    /**
+     * Generates an Excel spreadsheet containing all the applications stored in the repository.
+     * @return an Excel spreadsheet.
+     * @throws IOException if the spreadsheet cannot be written to the buffer.
+     */
+    @Get("/excel")
+    @ExecuteOn(TaskExecutors.BLOCKING)
```
sdelamo: Unnecessary since the controller is annotated with `@ExecuteOn`

starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/ExcelGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+                column.setter.accept(app, cell);
+                Optional.ofNullable(column.styler.apply(workbook)).ifPresent(cell::setCellStyle);
+            }
+        }
+        return workbook;
+    }
+
+    private enum ExcelColumn {
+        ID("id", CellType.NUMERIC, numeric(Application::getId)),
+        FEATURES("features", CellType.STRING, features(Application::getFeatures)),
```
sdelamo: column order should be 

```
date_created
type
language
build_tool
test_framework
jdk_version
micronaut_version
id
features
```

starter-analytics-postgres/src/test/groovy/io/micronaut/starter/analytics/postgres/GenerateExcelSpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+
+import java.util.concurrent.CompletableFuture
+
+import static io.micronaut.starter.analytics.postgres.ExcelGenerator.ExcelColumn.*
+
+@Property(name = "spec.name", value = "GenerateExcelSpec")
+@MicronautTest(
+        transactional = false,
+        environments = Environment.GOOGLE_COMPUTE)
+class GenerateExcelSpec extends Specification implements TestPropertyProvider {
```
sdelamo, line 31: annotate with `@spock.lang.Requires({ DockerClientFactory.instance().isDockerAvailable() })`

starter-analytics-postgres/src/test/groovy/io/micronaut/starter/analytics/postgres/GenerateExcelSpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+import spock.lang.Specification
+
+import java.util.concurrent.CompletableFuture
+
+import static io.micronaut.starter.analytics.postgres.ExcelGenerator.ExcelColumn.*
+
+@Property(name = "spec.name", value = "GenerateExcelSpec")
+@MicronautTest(
+        transactional = false,
+        environments = Environment.GOOGLE_COMPUTE)
```
sdelamo: why do we need to be in GCP env?

sdelamo (changes requested):

starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/ExcelGenerator.java
guillermocalvo: @sdelamo I just made the changes you requested. Please let me know if you want to go back to the previous [column definition](https://github.com/micronaut-projects/micronaut-starter/issues/1976#issue-1817886739):

> it should have a header row
>
> `id | features| type | language | buildTool | testFramework | jdkVersion | micronautVersion | dateCreated`

starter-analytics-postgres/src/test/groovy/io/micronaut/starter/analytics/postgres/GenerateExcelSpec.groovy
guillermocalvo: We don't. I had copied that part from an existing test.

I just removed env config from both of them.

starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/ExcelGenerator.java
sdelamo: I changed my mind from the original definition. I want features to be last in case I want to expand it immediately after

starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/AnalyticsController.java
```diff
@@ -93,4 +103,20 @@ HttpStatus applicationGenerated(@NonNull @Body Generated generated) {
...
+    /**
+     * Generates an Excel spreadsheet containing all the applications stored in the repository.
+     * @return an Excel spreadsheet.
+     * @throws IOException if the spreadsheet cannot be written to the buffer.
+     */
+    @Get("/excel")
+    @Produces(MediaType.MICROSOFT_EXCEL_OPEN_XML)
+    HttpResponse<StreamedFile> generateExcel() throws IOException {
+        try (ByteArrayOutputStream out = new ByteArrayOutputStream()) {
+            excelGenerator.generateExcel(applicationRepository.list(Pageable.UNPAGED)).write(out);
```
sdelamo: Should we change this code so we don't load every row into memory? Instead, we could paginate all the rows and build the Excel file contents with 50 rows each time.

The goal is to download an Excel file with every row. But I want us to be conscious about memory. We are running in 500 mb in cloud run. It is okay if it is slower. But I don't want it to crash.

Would not be better to do 100 queries to bring 50 rows from the database at a time and build the excel contents slowly instead of doing a single query, and bringing every entry into memory?.

sdelamo (changes requested):

starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/AnalyticsController.java
guillermocalvo: I just pushed some changes to make page size configurable via `excel.fetch.page.size` -- 50 by default.

starter-analytics-postgres/src/main/java/io/micronaut/starter/analytics/postgres/ExcelGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+    };
+
+    private static final Function<Workbook, CellStyle> CELL_STYLE_NULL = workbook -> null;
+
+    private final ApplicationRepository applicationRepository;
+    private final int pageSize;
+
+    public ExcelGenerator(
+            ApplicationRepository applicationRepository,
+            @Value("${excel.fetch.page.size:50}") int pageSize) {
```
sdelamo: Please, use a `@ConfigurationProperties` class not `@Value`.

sdelamo (changes requested):
Can you set `excel.fetch.page.size` in the test to a small number? I want the tests to trigger the pagination logic. 

E.g. 12 entries in the db and a pagination of 5.

sdelamo (approved):

PR #1985, opened 2023-07-28
Simplify test

sdelamo:
What do you think about this?


guillermocalvo (approved):
I'm not 100% sure if I prefer to manually start/stop two different application contexts instead of annotating the spec with `@MicronautTest`, but I'm good if we want to test this scenario like this 👍

PR #1998, opened 2023-08-03
Add missing test runtime dependencies for Neo4JBolt feature.

### LINKED ISSUE #1997: Neo4jBolt feature for Micronaut 4 missing runtime test dependencies
As per the [`micronaut-neo4j` docs](https://micronaut-projects.github.io/micronaut-neo4j/latest/guide/#breaks)

> Neo4J Harness 5.x adds compatibility for Java 17. Starting with Micronaut Framework 4.0.0 the Neo4j Harness requires a runtime dependency on `javax.annotation` and `javax.validation`
> 
> `testRuntimeOnly("javax.annotation:javax.annotation-api:1.3.2")`
> `testRuntimeOnly("javax.validation:validation-api:2.0.1.Final")`
> 

The `neo4j-bolt` feature should add these dependencies.

wetted:
closes #1997

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/database/Neo4jBolt.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/database/Neo4jBolt.java
@@ -32,6 +32,19 @@ public class Neo4jBolt implements Feature {
             .testRuntime()
             .build();
 
+    private static final Dependency DEPENDENCY_JAVAX_ANNOTATION_API = Dependency.builder()
+            .groupId("javax.annotation")
+            .artifactId("javax.annotation-api")
+            .testRuntime()
+            .build();
+
+    private static final Dependency DEPENDENCY_JAVAX_VALIDATION_API = Dependency.builder()
+            .groupId("javax.validation")
+            .artifactId("validation-api")
+            .version("2.0.1.Final")
+            .testRuntime()
+            .build();
+
     @Override
     @NonNull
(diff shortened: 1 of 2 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/database/Neo4jBolt.java
```diff
@@ -32,6 +32,19 @@ public class Neo4jBolt implements Feature {
...
+    private static final Dependency DEPENDENCY_JAVAX_ANNOTATION_API = Dependency.builder()
+            .groupId("javax.annotation")
+            .artifactId("javax.annotation-api")
+            .testRuntime()
+            .build();
+
+    private static final Dependency DEPENDENCY_JAVAX_VALIDATION_API = Dependency.builder()
+            .groupId("javax.validation")
+            .artifactId("validation-api")
+            .version("2.0.1.Final")
```
sdelamo, line 44: we need to set the version in the catalogue

sdelamo (changes requested):

graemerocher:
not sure if we should do this tbh, we should be using test containers for neo4j and remove the test harness all together

sdelamo:
> not sure if we should do this tbh, we should be using test containers for neo4j and remove the test harness all together

I agree. @wetted create an issue in the Neo4 repo to migrate to test containers.

PR #2025, opened 2023-08-23
Stop hardcoding a version of micronaut-security-aot

timyates:
We were hardcoding a version of micronaut-security-aot, which meant that we ended up with two versions of a core library when trying to run the AOT analysis.

This led to multiple beans being in the context as seen in

https://github.com/micronaut-projects/micronaut-security/issues/1420

The fix here imports the micronaut-platform for the aotPlugin configuration

I don't believe we need this for the maven dependency.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
@@ -6,7 +6,7 @@
                   <dependency>
                       <groupId>@(aotDependency.getGroupId())</groupId>
                       <artifactId>@(aotDependency.getArtifactId())</artifactId>
-                      <version>${micronaut.security.version}</version>
+                      <version>@(aotDependency.getVersion())</version>
                   </dependency>
 }
               </aotDependencies>
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
@@ -6,7 +6,6 @@
                   <dependency>
                       <groupId>@(aotDependency.getGroupId())</groupId>
                       <artifactId>@(aotDependency.getArtifactId())</artifactId>
-                      <version>@(aotDependency.getVersion())</version>
                   </dependency>
 }
               </aotDependencies>
--- a/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
@@ -6,6 +6,7 @@
                   <dependency>
                       <groupId>@(aotDependency.getGroupId())</groupId>
                       <artifactId>@(aotDependency.getArtifactId())</artifactId>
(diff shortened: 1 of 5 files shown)


timyates:
@alvarosanchez Can you check the maven change?  We don't need a version in the pom do we?

we end up with

```
      <plugin>
        <groupId>io.micronaut.maven</groupId>
        <artifactId>micronaut-maven-plugin</artifactId>
          <configuration>
              <aotDependencies>
                  <dependency>
                      <groupId>io.micronaut.security</groupId>
                      <artifactId>micronaut-security-aot</artifactId>
                  </dependency>
              </aotDependencies>
            <configFile>aot-${packaging}.properties</configFile>
          </configuration>
      </plugin>
```

alvarosanchez:
Looking at the plugin sources it appears to be like that, but better give it a try first. For some reason, I have explicit versions in the sample documentation, so I'm not really sure 😅

sdelamo (approved):
I have changed to use `BuildTestVerifier`

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
```diff
@@ -6,7 +6,6 @@
                   <dependency>
                       <groupId>@(aotDependency.getGroupId())</groupId>
                       <artifactId>@(aotDependency.getArtifactId())</artifactId>
-                      <version>@(aotDependency.getVersion())</version>
```
sdelamo: This should be `<version>${micronaut.security.version}</version>`

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
timyates: Ahhhh so then it comes from platform parent which is similar to what we're doing with Gradle

timyates: Done in https://github.com/micronaut-projects/micronaut-starter/pull/2025/commits/ba9a1f9035c39b2fd70b4b5b9017bb650ced1f73

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
```diff
@@ -6,6 +6,7 @@
                   <dependency>
                       <groupId>@(aotDependency.getGroupId())</groupId>
                       <artifactId>@(aotDependency.getArtifactId())</artifactId>
+                      <version>${micronaut.security.version}</version>
```
sdelamo: we have to pass this as a parameter to the rocker template. If we have another module contribute an AOT optimisations. e.g. aws, we will need `${micronaut.aws.version}`.

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/aotDependencies.rocker.raw
timyates: 🤦 

Done in https://github.com/micronaut-projects/micronaut-starter/pull/2025/commits/70d61fddf7db919146ce062f81612ccc72bb9a8d

sdelamo (approved):

PR #2032, opened 2023-08-31
Add features for Groovy optional modules

### LINKED ISSUE #2008: Add features for Groovy optional modules
@paulk-asert

> Also, would it make sense to have the Groovy optional modules as features that could be added by launch? E.g. The "groovy-datetime" would add the "groovy-datetime" module. Can this be done just for when the Groovy language is selected? Probably doesn't make sense to use those from other languages, even though it is just a Java library in this case for instance.

wetted:
closes #2008


sdelamo (changes requested):
Please create better descriptions and titles. I don't like the "Groovy Optional". Read the modules documentation and provide specific title and description for each dependency.

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/module/DateUtilGroovyModule.java
```diff
@@ -0,0 +1,45 @@
...
+import io.micronaut.core.annotation.NonNull;
+import jakarta.inject.Singleton;
+
+@Singleton
+public class DateUtilGroovyModule implements GroovyModuleFeature {
+
+    @Override
+    public String getName() {
+        return "groovy-dateutil";
+    }
```
timyates: Could we use a NAME constant?

```suggestion
    public static final String NAME = "groovy-dateutil";

    @Override
    public String getName() {
        return NAME;
    }
```

(here and in the other features)

starter-core/src/test/groovy/io/micronaut/starter/feature/lang/groovy/module/GroovyModuleFeatureSpec.groovy
```diff
@@ -0,0 +1,79 @@
...
+    void "test feature #moduleFeature with build tool #buildTool and Groovy language"(String moduleFeature, BuildTool buildTool) {
+        when:
+        String template = new BuildBuilder(beanContext, buildTool)
+                .language(Language.GROOVY)
+                .features([moduleFeature])
+                .render()
+        BuildTestVerifier verifier = BuildTestUtil.verifier(buildTool, template)
+
+        then:
+        verifier.hasDependency("org.apache.groovy", moduleFeature, Scope.TEST)
```
timyates: If it's a Groovy application, shouldn't this be an implementation dependency, not a `TEST` one?

timyates (changes requested):

starter-core/src/test/groovy/io/micronaut/starter/feature/lang/groovy/module/GroovyModuleFeatureSpec.groovy
wetted: The issue and @sdelamo otherwise both suggest adding them to the test classpath.

timyates: I don't think that's right... (and can't find a discussion around this on the issue) 🤔

If I have created an application in the Groovy language and I add the `groovy-json` feature, I'd think it's more normal that I want to parse json in my application code, rather than just in the tests

I'm willing to be overridden by @sdelamo or @paulk-asert if that wasn't the original intention though 😀

wetted: I have no preference/objection one way or the other. Sergio suggested it in a private discussion, "probably adding those dependencies to the test classpath". We can ask when he's back.

wetted: @timyates I changed to what I think is the correct compromise: if language is Groovy, add to compile classpath; otherwise, if test framework is Spock add to test classpath.

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/module/GroovyModuleFeature.java
```diff
@@ -0,0 +1,66 @@
...
+        Dependency.Builder builder = Dependency.builder()
+                .groupId("org.apache.groovy")
+                .artifactId(getName());
+        if (generatorContext.getLanguage() == Language.GROOVY) {
+            // if language is Groovy, add to compile classpath
+            generatorContext.addDependency(builder.compile());
+        } else if (generatorContext.getTestFramework() == TestFramework.SPOCK) {
+            // if language is not Groovy, but test if Spock, add to test classpath
+            generatorContext.addDependency(builder.test());
+        }
```
timyates: Is this better?  We have the validator to stop it being done?

```suggestion
        Dependency.Builder builder = Dependency.builder()
                .groupId(Groovy.GROUP_ID_GROOVY)
                .artifactId(getName());

        if (generatorContext.getLanguage() == Language.GROOVY) {
            // if language is Groovy, add to compile classpath
            builder.compile();
        } else {
            // otherwise add to test classpath
            builder.test();
        }
        generatorContext.addDependency(builder);
```

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/module/GroovyModuleFeature.java
```diff
@@ -0,0 +1,66 @@
...
+    default boolean supports(ApplicationType applicationType) {
+        return true;
+    }
+
+    @Override
+    @NonNull
+    default String getThirdPartyDocumentation() {
+        return "https://docs.groovy-lang.org/docs/latest/html/documentation/";
+    }
+
```
timyates: This is done in the feature classes themselves, and isn't required here

timyates (approved):
suggestion and un-required default method but otherwise 😎👍 

I think it makes much more sense for compile scope if you're creating a groovy app

sdelamo (approved):

sdelamo:
@paulk-asert what do you think?

starter-core/src/main/java/io/micronaut/starter/feature/Category.java
```diff
@@ -33,6 +33,7 @@ public class Category {
     public static final String DEV_TOOLS            = "Development Tools";
     public static final String DISTRIBUTED_CONFIG   = "Distributed Configuration";
     public static final String DOCUMENTATION        = "Documentation";
+    public static final String GROOVY_MODULE        = "Groovy Optional Module";
```
paulk-asert: I'd probably pluralize "Module" to "Modules".

paulk-asert:
> @paulk-asert what do you think?

This looks good to me. There is no perfect fit for languages/testing framework. I can use various extension methods from a Java application just as I would use static methods from commons-lang or guava. Similarly, I can write my JUnit tests in Groovy. But what is covered here is the sweet spot and will cover 90+% of the commons cases. Folks who don't want the common cases can always add in their own dependencies manually which is what they would currently do.

starter-core/src/main/java/io/micronaut/starter/feature/lang/groovy/module/GroovyModuleFeature.java
```diff
@@ -0,0 +1,60 @@
...
+import io.micronaut.starter.application.generator.GeneratorContext;
+import io.micronaut.starter.build.dependencies.Dependency;
+import io.micronaut.starter.feature.Category;
+import io.micronaut.starter.feature.Feature;
+import io.micronaut.starter.options.Language;
+
+/**
+ * Implementation note: GroovyModuleFeature is not a LanguageSpecificFeature,
+ * because it's perfectly reasonable to use for Spock framework, and the
+ * GroovyModuleFeature dependencies are added as test scope by default.
```
paulk-asert: I am not sure whether everything after the "and" is a legacy comment but I'd probably change "and" to "in which case" and get rid of the "by default" at the end.

sdelamo (approved):

PR #2039, opened 2023-09-08
Adding projectGroupId to micronaut starter

YasserLahrach:
Adding projectGroupId to micronaut starter

### DIFF
--- a/buildSrc/src/main/groovy/io.micronaut.internal.starter.base-module.gradle
+++ b/buildSrc/src/main/groovy/io.micronaut.internal.starter.base-module.gradle
@@ -8,7 +8,7 @@ repositories {
 }
 
 version project.projectVersion
-group "io.micronaut.starter"
+group projectGroup
 ext {
     isBuildSnapshot = version.toString().endsWith("-SNAPSHOT")
 }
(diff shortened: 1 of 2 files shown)


sdelamo:
why do we need this?

sdelamo (changes requested):
[group id is `io.micronaut.starter`](https://github.com/micronaut-projects/micronaut-starter/blob/4.1.x/buildSrc/src/main/groovy/io.micronaut.internal.starter.base-module.gradle#L11). Why do we need this PR?

YasserLahrach:
We use that variable to generate artifacts paths. so we had to use a workaround for micronaut starter project.

YasserLahrach:
We use that variable to generate artifacts paths. We had to use a workaround for micronaut starter project.

sdelamo:
why do you need to generate artifacts paths? What are you doing with starter?.

sdelamo (changes requested):
We should not duplicate the project group definition https://github.com/micronaut-projects/micronaut-starter/blob/4.1.x/buildSrc/src/main/groovy/io.micronaut.internal.starter.base-module.gradle#L11

alvarosanchez (approved):

sdelamo:
@YasserLahrach @alvarosanchez can you please comment on 

> why do you need to generate artifacts paths? What are you doing with starter?.

alvarosanchez:
Starter is used for https://www.graal.cloud/gcn/launcher/

PR #2047, opened 2023-09-12
AWS Lambda and CDK configuration with Java 17 shouldn't set JAVA_TOOL_OPTIONS variable

### LINKED ISSUE #2043: AWS Lambda and CDK configuration with Java 17
Thanks for reporting an issue, please review the task list below before submitting the
issue. Your issue report will be closed if the issue is incomplete and the below tasks not completed.

NOTE: If you are unsure about something and the issue is more of a question a better place to ask questions is on Stack Overflow (http://stackoverflow.com/tags/micronaut) or Gitter (https://gitter.im/micronautfw/). DO NOT use the issue tracker to ask questions.

### Description

The Java 17 release of the Lambda managed runtime has changed the default value of tiered compilation. It doesn't need to be set in CDK when you create a Lambda function.

### Task List

- [x] Steps to reproduce provided
- [x] Stacktrace (if present) provided
- [ ] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. mn create-app --build=maven --jdk=17 --lang=java --test=junit --features=aws-cdk,aws-lambda com.example.demo

### Expected Behaviour

When using Java 17 the following environment variable shouldn't be created.

```
Map<String, String> environmentVariables = new HashMap<>();
        // https://aws.amazon.com/blogs/compute/optimizing-aws-lambda-function-performance-for-java/
        environmentVariables.put("JAVA_TOOL_OPTIONS", "-XX:+TieredCompilation -XX:TieredStopAtLevel=1");
```

### Actual Behaviour

It is being set.

### Environment Information

- **Operating System**: https://micronaut.io/launch
- **Micronaut Version:** Tested with 4.1 on [launch.micronaut.io](https://micronaut.io/launch)
- **JDK Version:** 17 https://micronaut.io/launch

### Example Application

- [TODO: link to github repository with example that reproduces the issue](https://micronaut.io/launch)

wetted:
fixes #2043

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/aws/template/cdkappstack.rocker.raw
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/aws/template/cdkappstack.rocker.raw
@@ -138,10 +138,6 @@ public class AppStack extends Stack {
                 .build());
 }
 @if (functionId != null) {
-@if (!nativeImage) {
-        // https://aws.amazon.com/blogs/compute/optimizing-aws-lambda-function-performance-for-java/
-        environmentVariables.put("JAVA_TOOL_OPTIONS", "-XX:+TieredCompilation -XX:TieredStopAtLevel=1");
-}
 @if (applicationType == ApplicationType.DEFAULT) {
         Function function = MicronautFunction.create(ApplicationType.DEFAULT,
 }


sdelamo (approved):

msailes:
I think you want to keep the condition for native image. Something like

if not (native image or java 17) 

For other versions of Java (8, 11) this still applies.

sdelamo:
> I think you want to keep the condition for native image. Something like
> 
> if not (native image or java 17)
> 
> For other versions of Java (8, 11) this still applies.

we don't support java 8 and 11 for Micronaut Framework 4. Thus, I think we can remove this.

msailes:
👏🏻  of course!

PR #2052, opened 2023-09-14
Exclude dependencies in Maven for metrics annotation

timyates:
The metrics-annotation has an api dependency on aop, which is pulling in other versions of things.

This PR excludes `micronaut-core` from the dependencies for the metrics annotation processor.

Excluding `micronaut-inject` as with the other processors is not enough as we also need to exclude `io.micronaut.core.beans.BeanIntrospection$Builder`


alvarosanchez (approved):

sdelamo (commented):
> The metrics-annotation has an api dependency on aop, which is pulling in other versions of things.

should not we change the `api` scope?

sdelamo (approved):

PR #2054, opened 2023-09-14
Exclude micronaut-core from microstream-annotation

timyates:
As with https://github.com/micronaut-projects/micronaut-starter/pull/2052 we need to add an exclusion to prevent different versions of micronaut-core ending up in the annotation classpath


sdelamo (commented):
Do we have a test that reproduces that this is an issue?

timyates:
@sdelamo we do now 😜

sdelamo (approved):

PR #2064, opened 2023-09-20
Fix MySQL dependency

burtbeckwith:
[WARNING] The artifact mysql:mysql-connector-java:jar:8.0.33 has been relocated to com.mysql:mysql-connector-j:jar:8.0.33: MySQL Connector/J artifacts moved to reverse-DNS compliant Maven 2+ coordinates.


sdelamo (approved):
fixed the checkstyle  https://github.com/micronaut-projects/micronaut-starter/pull/2064/commits/a24673735a17f3e79ae445a2d765138256dea161

sdelamo (changes requested):
There are tests failures. https://ge.micronaut.io/s/kdtegh4hbp4bk/tests/overview?outcome=FAILED

@burtbeckwith could you look into it?

sdelamo:
@wetted can you look into this?

wetted:
ping @sdelamo  @burtbeckwith 

The problem with this PR (i.e. the test failures) is that `mysql:mysql-connector-java` is exposed as a managed dependency by `micronaut-sql` but with this PR the coordinates change to an unmanaged dependency `com.mysql:mysql-connector-j`, but this PR treats it as managed. I can fix this PR easily enough by adding the new dependency as an unmanaged dependency. The tests fail because the dependency isn't qualified with a version (i.e. "8.1.0").

I think the correct thing to do is to change the managed dependency in `micronaut-sql` to the new coordinates, and hold this PR until it can point to that update.

I can work on either or both. Let me know how I should proceed. In any case this PR needs to point to the next minor and not a patch release.

sdelamo:
@wetted I started a PR in SQL https://github.com/micronaut-projects/micronaut-sql/pull/1106

wetted:
ping @sdelamo this is now good to go

sdelamo (approved):

PR #2069, opened 2023-09-25
Gradle plugin 4.0.4

timyates:
Contains the fixes to the CRaC docker tasks


sdelamo (changes requested):
we have to change it everywhere https://github.com/micronaut-projects/micronaut-starter/commit/dccb255e49dd857710337debcf77f879cebd9ed2

timyates:
Instead, I've removed it from settings.gradle (so we only have to change it in one place going forward)

...assuming it works

sdelamo (approved):

PR #2079, opened 2023-09-27
Community feature for slf4j-simple-logger 

### LINKED ISSUE #1654: Considering adding a community feature for slf4j-simple-logger
https://github.com/GoodforGod/slf4j-simple-logger

wetted:
https://github.com/GoodforGod/slf4j-simple-logger

closes #1654

### DIFF
--- a/starter-api/src/test/groovy/io/micronaut/starter/api/FeatureControllerSpec.groovy
+++ b/starter-api/src/test/groovy/io/micronaut/starter/api/FeatureControllerSpec.groovy
@@ -42,6 +42,7 @@ class FeatureControllerSpec extends Specification {
                 'agorapulse-micronaut-slack',
                 'agorapulse-micronaut-worker',
                 'camunda-zeebe',
+                'goodforgod-slf4j-simple-logger'
         ].sort()
     }
 
--- a/starter-api/src/test/groovy/io/micronaut/starter/api/FeatureControllerSpec.groovy
+++ b/starter-api/src/test/groovy/io/micronaut/starter/api/FeatureControllerSpec.groovy
@@ -41,8 +41,7 @@ class FeatureControllerSpec extends Specification {
                 'agorapulse-micronaut-permissions',
                 'agorapulse-micronaut-slack',
                 'agorapulse-micronaut-worker',
-                'camunda-zeebe',
-                'goodforgod-slf4j-simple-logger'
+                'camunda-zeebe'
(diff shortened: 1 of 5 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/logging/Slf4jSimpleLogger.java
```diff
@@ -0,0 +1,75 @@
...
+
+@Singleton
+public class Slf4jSimpleLogger implements CommunityFeature {
+    public static final String NAME = "goodforgod-slf4j-simple-logger";
+
+    private static final String SLF4J_SIMPLE_LOGGER_ARTIFACT_ID = "slf4j-simple-logger";
+
+    private static final Dependency SLF4J_SIMPLE_LOGGER_DEPENDENCY = Dependency.builder()
+            .lookupArtifactId(SLF4J_SIMPLE_LOGGER_ARTIFACT_ID)
+            .compile()
```
sdelamo: This should be `runtimeOnly`

sdelamo (changes requested):
Ping @GoodforGod

sdelamo (changes requested):
Thinking about this. I think we should create a OneOf marker interface for Slf4j implementation. Users should not add this dependency and logback at the same time. They should add either this one or logback.

wetted:
> Thinking about this. I think we should create a OneOf marker interface for Slf4j implementation. Users should not add this dependency and logback at the same time. They should add either this one or logback.

@sdelamo I already changed the `Slf4jSimpleLogger` feature to be a `LoggingFeature`. I believe that satisfies what you want since the other logger features are `LoggingFeature`, and that is a `OneOfFeature`.

Or did I misunderstand?

starter-api/src/test/groovy/io/micronaut/starter/api/FeatureControllerSpec.groovy
```diff
@@ -41,7 +41,7 @@ class FeatureControllerSpec extends Specification {
                 'agorapulse-micronaut-permissions',
                 'agorapulse-micronaut-slack',
                 'agorapulse-micronaut-worker',
-                'camunda-zeebe',
```
sdelamo, line 44: trailing comma is also allowed in groovy lists

sdelamo (approved):

PR #2080, opened 2023-09-28
Switch to Gradle-Kotlin as the default build tool

### LINKED ISSUE #2082: Make the Gradle Kotlin DSL the default
### Discussed in https://github.com/micronaut-projects/micronaut-starter/discussions/2062

<div type='discussions-op-text'>

<sup>Originally posted by **graemerocher** September 18, 2023</sup>
We should make the Gradle Kotlin DSL the default because:

a. It is now the default in Gradle
b. It doesn't break for each release of the JDK
c. It provides a better IDE experience</div>

timyates:
See: https://github.com/micronaut-projects/micronaut-starter/discussions/2062

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/coherence/CoherenceFeature.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/coherence/CoherenceFeature.java
@@ -18,9 +18,9 @@ package io.micronaut.starter.feature.coherence;
 import io.micronaut.starter.application.ApplicationType;
 import io.micronaut.starter.application.generator.GeneratorContext;
 import io.micronaut.starter.build.dependencies.Dependency;
+import io.micronaut.starter.build.dependencies.MicronautDependencyUtils;
 import io.micronaut.starter.feature.Category;
 import io.micronaut.starter.feature.Feature;
-import io.micronaut.starter.options.BuildTool;
 import jakarta.inject.Singleton;
 
 /**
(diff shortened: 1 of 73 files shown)


sdelamo (approved):
I wonder if we have test coverage which tests everything we generate it is Gradle Kotlin DSL compatible.

timyates:
> I wonder if we have test coverage which tests everything we generate it is Gradle Kotlin DSL compatible.

Think it's worth spending some time looking for `BuildTool.GRADLE` only tests (with missing `BuildTool.GRADLE_KOTLIN`)?

timyates:
I'm going to mark this as pending whilst I work on the coverage in the test-feature modules

starter-core/src/main/java/io/micronaut/starter/feature/coherence/CoherenceFeature.java
```diff
@@ -32,11 +32,11 @@
 @Singleton
 public class CoherenceFeature implements Feature {
 
-    public static final String COHERENCE_VERSION = "21.06.2";
```
timyates, line 35: Previously we were adding this version to `gradle.properties`, and then using that as the version in the Gradle and Maven builds.

Now the Coherence module has a BOM, I believe this is the wrong behaviour (and it would require changes to the kotlin dsl scripts to import the property)

I've removed this from here, and the version of the `com.oracle.coherence.ce` artifacts instead comes from the coherence bom

timyates:
Hmmm...rocker config needs fixing too I think...

starter-core/src/main/java/io/micronaut/starter/feature/view/gradlePluginRocker.rocker.raw
timyates, line 1: This works for both the Groovy and Kotlin DSL

PR #2098, opened 2023-10-02
build: Check via gradle tasks versions are not snapshots

sdelamo:
This PR adds Gradle Tasks to ensure a release can be triggered with non snapshot versions of `projectVersion`, `micronautCoreVersion` and micronautVersion`.

These tasks help to have a release with integrity.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,18 +20,18 @@ jobs:
       - name: Set the current release version
         id: release_version
         run: echo "release_version=${GITHUB_REF:11}" >> $GITHUB_OUTPUT
-      - name: projectVersion is not a snapshot
-        run: ./gradlew projectVersionNotSnapshot
-      - name: Micronaut Core is not a snapshot
-        run: ./gradlew micronautCoreNotSnapshot
-      - name: Micronaut Platform is not a snapshot
-        run: ./gradlew micronautPlatformNotSnapshot
       - name: Run pre-release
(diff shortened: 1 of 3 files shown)


melix (approved):

.github/workflows/release.yml
```diff
@@ -20,6 +20,12 @@ jobs:
       - name: Set the current release version
         id: release_version
         run: echo "release_version=${GITHUB_REF:11}" >> $GITHUB_OUTPUT
+      - name: projectVersion is not a snapshot
+        run: ./gradlew projectVersionNotSnapshot
+      - name: Micronaut Core is not a snapshot
+        run: ./gradlew micronautCoreNotSnapshot
+      - name: Micronaut Platform is not a snapshot
+        run: ./gradlew micronautPlatformNotSnapshot
```
timyates: I'm probably wrong, but isn't it the pre-release action that sets these versions to non-snapshot versions?

So shouldn't these checks come after the `Run pre-release` step?

sdelamo: `micronautCoreVersion` and `micronautPlatformVersion` are set manually. I move these checks after the `pre-release` step.

timyates (approved):

.github/workflows/release.yml
timyates: Cool (projectVersion is not set manually I assume?)

PR #2106, opened 2023-10-03
Fix Swagger

timyates:
We seem to be missing the static resource mapping for swagger and rapidoc.

This change adds the 2 static-resource mappings in to the config so they work again.

Run locally with `./gradlew :s-w-n:run` and then navigate to

http://localhost:8080/rapidoc/index.html
http://localhost:8080/swagger-ui/index.html


sdelamo (changes requested):
please, add tests

sdelamo (approved):

graemerocher (approved):

PR #2108, opened 2023-10-03
Add architecture as matrix for OSX snapshot action

timyates:


timyates:
Can't run on the aarch64 runners.  They only exist as Xlarge runners, so cost money

PR #2142, opened 2023-10-19
Add Java 21 as a JDK to starter

timyates:


starter-core/src/main/java/io/micronaut/starter/feature/function/azure/AzureFunctionFeatureValidator.java
```diff
@@ -0,0 +1,46 @@
...
+import io.micronaut.starter.options.Options;
+import jakarta.inject.Singleton;
+
+import java.util.Set;
+
+@Singleton
+public class AzureFunctionFeatureValidator implements FeatureValidator {
+
+    public static boolean supports(JdkVersion jdkVersion) {
+        return jdkVersion == JdkVersion.JDK_8 || jdkVersion == JdkVersion.JDK_11 || jdkVersion == JdkVersion.JDK_17;
```
sdelamo, line 31: I think we should remove 8 and 11

starter-core/src/main/java/io/micronaut/starter/feature/function/gcp/GoogleCloudFunctionFeatureValidator.java
```diff
@@ -28,24 +28,28 @@
 @Singleton
 public class GoogleCloudFunctionFeatureValidator implements FeatureValidator {
 
+    private static boolean supports(JdkVersion jdkVersion) {
+        return jdkVersion == JdkVersion.JDK_11 || jdkVersion == JdkVersion.JDK_17;
```
sdelamo, line 32: we should remove 11

sdelamo (approved):

PR #2096, opened 2023-10-02
regression: Enable groovy tests

sdelamo:
We have a regression in groovy support from 4.1.2 to 4.1.3.

Core changes:

https://github.com/micronaut-projects/micronaut-core/compare/v4.1.6...v4.1.8

Starter changes: 

https://github.com/micronaut-projects/micronaut-starter/compare/v4.1.2...v4.1.3

 This PR reverts commit a53a2d7a3138a8795706a3d1ff0d5b5827047d35.

PR #2150, opened 2023-10-31
Add compileOnly http-client dependency with Groovy and Gradle

timyates:
http-client has been moved to testImplementation.

But micronaut-http-client is required by micronaut-http-validation, which is a compileOnly dependency with Groovy projects.

So Gradle builds fail in guides

https://ge.micronaut.io/s/hspmjzngyrw4i/console-log?page=1#L57


sdelamo (changes requested):
It misses tests

timyates:
Tests added and fixed the issue that showed up...

Sorry about that, and thanks!

sdelamo (changes requested):
Please, change the PR base to 4.1.x

timyates:
Ok...  so after investigation, we [already fixed this](https://github.com/micronaut-projects/micronaut-starter/pull/2096) on 4.1.x, but it's not been merged up to 4.2.x

I'll close this PR, and work on the merge

PR #2156, opened 2023-11-01
Configure compileTestResourcesKotlin when using Kotest or Kotlin

timyates:
Previously we configured compileKotlin and compileTestKotlin when using Kotest or Kotlin and Java 21.

This failed with one of the guides when generated with Java 21 https://ge.micronaut.io/s/kbdu4j2l3qg44/console-log?page=2#L1057

This is because we were not comfiguring kotlinTestResources as well.


timyates:
Not sure this is right :-/

PR #2171, opened 2023-11-09
feat: add buildless feature

sgammon:
This PR adds support for [Buildless](https://less.build) in the project generator. Previously discussed with @sdelamo. As a precursor, I've also added support for `settings.gradle.kts` imports, which need to be hoisted to the top of the file.

- [x] feat: add support for settings-time imports in gradle
- [x] feat: add buildless feature

Screenshot 2023-11-09 at 12 06 06 AM

Demo of a generated project:
[demo-buildless.zip](https://github.com/micronaut-projects/micronaut-starter/files/13305607/demo-buildless.zip)

Most notably, in **`settings.gradle.kts`**:
```kotlin
plugins {
    id("build.less") version("1.0.0-beta1")
}

buildless {
    // this call is enough to configure the cache, although, remember to
    // set BUILDLESS_APIKEY in your environment.
}
```

And, in **`gradle.properties`**:
```properties
org.gradle.caching=true
```


starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
```diff
@@ -0,0 +1,134 @@
...
+import io.micronaut.starter.feature.build.gradle.templates.gradleBuildCache;
+import io.micronaut.starter.template.RockerWritable;
+import jakarta.inject.Singleton;
+
+@Singleton
+public class Buildless implements Feature, BuildCacheConfiguration {
+    public static final boolean BUILDLESS_ENABLE_PLUGIN = true;
+    public static final String FEATURE_NAME_BUILDLESS = "buildless";
+    public static final String BUILDLESS_PLUGIN_ID = "build.less";
+    public static final String BUILDLESS_PLUGIN_VERSION = "1.0.0-beta1";
```
sdelamo: this should be defined in [tool file](https://github.com/micronaut-projects/micronaut-starter/blob/4.1.x/gradle/templates.versions.toml)

starter-core/src/main/java/io/micronaut/starter/feature/build/BuildCacheConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.starter.feature.build;
+
+import io.micronaut.core.annotation.Nullable;
+
+public interface BuildCacheConfiguration {
```
sdelamo: should not this be in the buildless package?

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
```diff
@@ -0,0 +1,134 @@
...
+import io.micronaut.starter.feature.Category;
+import io.micronaut.starter.feature.Feature;
+import io.micronaut.starter.feature.build.BuildCacheConfiguration;
+import io.micronaut.starter.feature.build.gradle.templates.buildlessGradlePlugin;
+import io.micronaut.starter.feature.build.gradle.templates.gradleBuildCache;
+import io.micronaut.starter.template.RockerWritable;
+import jakarta.inject.Singleton;
+
+@Singleton
+public class Buildless implements Feature, BuildCacheConfiguration {
```
sdelamo: this should implement `CommunityFeature` instead of `Feature`

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
```diff
@@ -30,11 +30,7 @@
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
```
timyates:
```suggestion
import java.util.Collections;
import java.util.HashSet;
import java.util.List;
import java.util.Objects;
import java.util.Set;
```

starter-core/src/main/java/io/micronaut/starter/build/gradle/GradleBuild.java
```diff
@@ -81,6 +77,11 @@ public List<GradlePlugin> getPlugins() {
         return plugins.stream().filter(gradlePlugin -> gradlePlugin.getGradleFile() == GradleFile.BUILD).collect(Collectors.toList());
     }
 
+    @NonNull
+    public List<String> getSettingsImports() {
+        return plugins.stream().filter(gradlePlugin -> gradlePlugin.getGradleFile() == GradleFile.SETTINGS).map(GradlePlugin::getSettingsImports).flatMap(Collection::stream).collect(Collectors.toList());
```
timyates:
```suggestion
        return plugins.stream().filter(gradlePlugin -> gradlePlugin.getGradleFile() == GradleFile.SETTINGS).map(GradlePlugin::getSettingsImports).flatMap(Collection::stream).toList();
```

Can use Java 17 `Stream.toList()`

timyates (changes requested):
Hiya 👋 

Cool!  We're missing tests.  We have tests which validate the contents of the generated project files ([such as this Gradle Enterprise feature test](https://github.com/micronaut-projects/micronaut-starter/blob/a2048d187a4bd24e38ee256d856f2277e27bd01e/starter-core/src/test/groovy/io/micronaut/starter/feature/build/GradleEnterpriseSpec.groovy)), and functional tests which test that a project including this feature can be compiled by the respective build tool ([as in this example for the http-client-jdk feature](https://github.com/micronaut-projects/micronaut-starter/blob/3c847dcdf0b408801a71324934dcea13d0d56d4f/test-features/src/test/groovy/io/micronaut/starter/core/test/feature/httpclient/HttpClientJdkSpec.groovy))

We also need to think what happens if this feature is combined with the Gradle Enterprise (Develocity) feature -- Especially in the case of Maven I guess where both will be writing to extensions.xml

### Edit

It's fine to not support Maven, but we will need a [feature validator](https://github.com/micronaut-projects/micronaut-starter/blob/a5f4cbd0420ee9774dfd12e8a36aad61f0b462d9/starter-core/src/main/java/io/micronaut/starter/feature/validation/FeatureValidator.java) which prevents people choosing this selection

sgammon:
Wow, thank you guys for the quick review! I'll address these comments. @sdelamo / @timyates

starter-core/src/main/java/io/micronaut/starter/feature/build/BuildCacheConfiguration.java
sgammon: @sdelamo this is a general configuration for any kind of Gradle build caching, following the structure of their [build cache configuration](https://docs.gradle.org/current/userguide/build_cache.html).

I figure if there are future providers, or some feature which needs to activate the build cache, it can use this configuration. The Buildless feature uses this configuration the same way. If you'd like I can move it to be next to the `Buildless` feature.

starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
sgammon: @sdelamo I implemented `CommunityFeature`, but this code prevents tests from running properly (`ENABLE_COMMUNITY_FEATURES` is currently `false`):

https://github.com/micronaut-projects/micronaut-starter/blob/9c68883693ea2f82cc1e4789c4d8571ebd8ad577/starter-core/src/main/java/io/micronaut/starter/feature/CommunityFeatureValidator.java#L34-L37

So I've flipped `ENABLE_COMMUNITY_FEATURES`:
https://github.com/micronaut-projects/micronaut-starter/blob/0f3839464278b8e03b9db227d14cae60a968514e/starter-core/src/main/java/io/micronaut/starter/feature/CommunityFeatureValidator.java#L27

Tests still pass as far as I can tell; is this the right thing to do?

sgammon:
@timyates / @sdelamo Okay, I've cleaned things up a bit and added tests. Let me address other comments here:

- I'm a bit worried about `ENABLE_COMMUNITY_FEATURES`. If it must remain `false`, and we must implement `CommunityFeature`, then things break. Looking for advice on this one

- **Tests:** I've added both types of tests; basic spec and a `test-features` spec. I've used the samples you provided, @timyates -- thank you. The plugin is really simple, so there isn't much to test here, and things seem to work. If you see anything I can improve in my tests please let me know.

- **Maven:** With regard to Maven support, thank you for catching that, we *do* support Maven, I'll add those changes on top of these shortly and include tests. Just want to make sure this passes tests / I'm on the right track before building more.

- **Develocity:** We use GE/Develocity in tandem with Buildless in our environment. I'll make sure there is no case where combining the two creates an invalid project structure. If there is a conflict, we are happy to yield to Develocity.

Obviously, once you guys approve, I will rebase so that it ends up in one clean commit.

timyates:
@sgammon Hiya!  Apologies, today got ahead of me and I've run out of time

- ENABLE_COMMUNITY_FEATURES: Yeah, when we moved to micronaut 4 community features that were compiled against Micronaut 3 would no longer work.  We have discussed this and are thinking of adding a "maxVersion" or something to community features. TBD 😄 

And don't worry about rebasing to a single commit, we usually "Squash and merge" in GitHub, so it becomes one commit then

I'll dig in to this properly on Monday, but a quick suggestion for the feature is:

```diff
diff --git a/starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java b/starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
index 13b36340d5..693f283013 100644
--- a/starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
@@ -21,7 +21,6 @@ import io.micronaut.starter.build.gradle.GradleFile;
 import io.micronaut.starter.build.gradle.GradlePlugin;
 import io.micronaut.starter.feature.Category;
 import io.micronaut.starter.feature.CommunityFeature;
-import io.micronaut.starter.feature.Feature;
 import io.micronaut.starter.feature.build.BuildCacheConfiguration;
 import io.micronaut.starter.feature.build.gradle.templates.buildlessGradlePlugin;
 import io.micronaut.starter.feature.build.gradle.templates.gradleBuildCache;
@@ -30,10 +29,11 @@ import jakarta.inject.Singleton;

 @Singleton
 public class Buildless implements CommunityFeature, BuildCacheConfiguration {
+
     public static final boolean BUILDLESS_ENABLE_PLUGIN = true;
     public static final String FEATURE_NAME_BUILDLESS = "buildless";
     public static final String BUILDLESS_PLUGIN_ID = "build.less";
-    public static final String BUILDLESS_PLUGIN_VERSION = "1.0.0-beta1";
+    public static final String BUILDLESS_PLUGIN_ARTIFACT = "buildless-plugin-gradle";

     @Override
     public String getName() {
@@ -114,7 +114,7 @@ public class Buildless implements CommunityFeature, BuildCacheConfiguration {
         var builder = GradlePlugin.builder()
                 .gradleFile(GradleFile.SETTINGS)
                 .id(BUILDLESS_PLUGIN_ID)
-                .version(BUILDLESS_PLUGIN_VERSION);
+                .lookupArtifactId(BUILDLESS_PLUGIN_ARTIFACT);

         if (!isUseCustomCachePlugin()) {
             // use a generic HTTPS endpoint for caching
```

This way, the feature version is looked up in the `templates.versions.toml` file so we don't have to worry about it being in many places

And if you agree, I think the `BuildlessSpec` in `starter-core` can be switched to:

```diff
diff --git a/starter-core/src/test/groovy/io/micronaut/starter/feature/buildless/BuildlessSpec.groovy b/starter-core/src/test/groovy/io/micronaut/starter/feature/buildless/BuildlessSpec.groovy
index 7267490e8c..6652396807 100644
--- a/starter-core/src/test/groovy/io/micronaut/starter/feature/buildless/BuildlessSpec.groovy
+++ b/starter-core/src/test/groovy/io/micronaut/starter/feature/buildless/BuildlessSpec.groovy
@@ -1,61 +1,36 @@
 package io.micronaut.starter.feature.buildless

 import io.micronaut.starter.ApplicationContextSpec
-import io.micronaut.starter.BuildBuilder
 import io.micronaut.starter.application.ApplicationType
-import io.micronaut.starter.application.Project
-import io.micronaut.starter.build.gradle.GradleBuild
-import io.micronaut.starter.build.gradle.GradleDsl
-import io.micronaut.starter.feature.build.MicronautGradleEnterprise
-import io.micronaut.starter.feature.build.gradle.templates.settingsGradle
+import io.micronaut.starter.build.dependencies.CoordinateResolver
 import io.micronaut.starter.fixture.CommandOutputFixture
 import io.micronaut.starter.options.BuildTool
 import io.micronaut.starter.options.Language
-import spock.lang.Subject
+import io.micronaut.starter.options.Options

 class BuildlessSpec extends ApplicationContextSpec implements CommandOutputFixture {

-    @Subject
-    Buildless buildless = beanContext.getBean(Buildless.class)
-
     @Override
     Map<String, Object> getConfiguration() {
         ["spec.name": "BuildlessSpec"]
     }

     void "if you add the buildless feature it is configured for #buildTool"() {
-        given:when:
-        BuildBuilder builder = new BuildBuilder(beanContext, buildTool)
-                    .language(Language.JAVA)
-                    .applicationType(ApplicationType.DEFAULT)
-                    .features(["buildless"])
-        Project project = builder.getProject()
-        GradleBuild gradleBuild = (GradleBuild) builder.build(false)
-        String settings = settingsGradle.template(project, gradleBuild, false, []).render().toString()
-
-        then: 'we we should have a plugins block'
-        settings.contains('plugins {')
-
-        and: 'buildless defaults should enable the plugin'
-        assert buildless.isUseCustomCachePlugin()
-        assert buildless.isEnableRemoteCache()
-        assert buildless.isRemoteCachePushEnabled()
-        assert buildless.isUseExpectContinue()
-        assert buildless.getRemoteCacheType() != null
-        assert buildless.getRemoteCacheUri() != null
-
-        and: 'buildless imports and block should be added'
-        assert settings.contains("buildless {")
-        if (gradleBuild.getDsl() == GradleDsl.KOTLIN) {
-            assert settings.contains("import build.less.plugin.settings.buildless")
-        }
-
-        and: 'buildless plugin should be added'
-        if (buildTool == BuildTool.GRADLE_KOTLIN) {
-            assert settings.contains('    id("build.less")')
-        } else {
-            assert settings.contains('    id "build.less"')
-        }
+        when:
+        def output = generate(ApplicationType.DEFAULT, new Options(Language.JAVA, buildTool), [Buildless.FEATURE_NAME_BUILDLESS])
+        def version = beanContext.getBean(CoordinateResolver).resolve(Buildless.BUILDLESS_PLUGIN_ARTIFACT).get().version
+        def settings = buildTool == BuildTool.GRADLE ? output['settings.gradle'] : output['settings.gradle.kts']
+        def expectedPlugin = buildTool == BuildTool.GRADLE ? """id "build.less" version "$version\"""" : """id("build.less") version("$version")"""
+
+        then:
+        settings.contains("    $expectedPlugin")
+
+        and:
+        settings.contains("""\
+buildless {
+  // this call is enough to configure the cache, although, remember to
+  // set BUILDLESS_APIKEY in your environment.
+}""")

         where:
         buildTool << BuildTool.valuesGradle()
```

So this uses the `generate` method to actually create the files via starter, and then checks the settings file that's generated to ensure the plugin and config are applied (and we fetch the version from the toml file again to save having to chenge it each release)

Hope this makes sense!

It's looking really nice 😄👍

starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
```diff
@@ -0,0 +1,144 @@
+/*
+ * Copyright 2017-2021 original authors
```
timyates: 2023

starter-core/src/main/java/io/micronaut/starter/feature/build/BuildCacheConfiguration.java
timyates: I'm not sure we need this in this PR...

The work to support swappable build caches is probably out of scope here

Unless I'm missing why this is here

starter-core/src/test/groovy/io/micronaut/starter/feature/buildless/BuildlessSpec.groovy
```diff
@@ -0,0 +1,68 @@
...
+import io.micronaut.starter.build.gradle.GradleBuild
+import io.micronaut.starter.build.gradle.GradleDsl
+import io.micronaut.starter.feature.build.MicronautGradleEnterprise
+import io.micronaut.starter.feature.build.gradle.templates.settingsGradle
+import io.micronaut.starter.fixture.CommandOutputFixture
+import io.micronaut.starter.options.BuildTool
+import io.micronaut.starter.options.Language
+import spock.lang.Subject
+
+class BuildlessSpec extends ApplicationContextSpec implements CommandOutputFixture {
```
timyates, line 12: [See my comment](https://github.com/micronaut-projects/micronaut-starter/pull/2171#issuecomment-1806191525) about how we can change this to check the settings file and pick up the version from the template.version.toml file

I tried to push, but can't make changes to this PR, but this is what I ended up with for a test here

```groovy
class BuildlessSpec extends ApplicationContextSpec implements CommandOutputFixture {

    void 'test README.md with buildless feature contains links to docs'() {
        when:
        def output = generate([Buildless.NAME])
        def readme = output["README.md"]

        then:
        readme.contains("[https://docs.less.build/](https://docs.less.build/)")
    }

    void "feature buildless properties validation"() {
        given:
        def feature = beanContext.getBean(Buildless)

        expect:
        for (applicationType in ApplicationType.values()) {
            assert feature.supports(applicationType)
        }

        feature.category == Category.DEV_TOOLS
        feature.name == Buildless.NAME
    }

    void "buildless configured correctly for #buildTool"() {
        when:
        def output = generate(ApplicationType.DEFAULT, new Options(Language.JAVA, buildTool), [Buildless.NAME])
        def version = beanContext.getBean(CoordinateResolver).resolve(Buildless.BUILDLESS_PLUGIN_ARTIFACT).get().version
        def settings = buildTool == BuildTool.GRADLE ? output['settings.gradle'] : output['settings.gradle.kts']
        def expectedPlugin = buildTool == BuildTool.GRADLE ? """id "build.less" version "$version\"""" : """id("build.less") version("$version")"""

        then: 'we have the plugin in the settings file'
        settings.contains("    $expectedPlugin")

        and: 'we have the configuration in the settings file'
        settings.contains("""\
buildless {
    // this call is enough to configure the cache, although, remember to
    // set BUILDLESS_APIKEY in your environment.
}""")

        and: 'caching is enabled in the properties file'
        output['gradle.properties'].contains("org.gradle.caching=true")

        where:
        buildTool << BuildTool.valuesGradle()
    }

    void "io.micronaut.starter.feature.buildless.Buildless is visible"() {
        expect:
        beanContext.getBean(Buildless).visible
    }
}
```

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/buildlessGradlePlugin.rocker.raw
```diff
@@ -0,0 +1,9 @@
+@import io.micronaut.starter.feature.build.BuildCacheConfiguration
+@args (BuildCacheConfiguration config)
+
+@if (config.isUseCustomCachePlugin()) {
+buildless {
+  // this call is enough to configure the cache, although, remember to
+  // set BUILDLESS_APIKEY in your environment.
```
timyates:
```suggestion
    // this call is enough to configure the cache, although, remember to
    // set BUILDLESS_APIKEY in your environment.
```

4 space indent as elsewhere

test-features/src/test/groovy/io/micronaut/starter/core/test/feature/buildless/BuildlessSpec.groovy
```diff
@@ -0,0 +1,27 @@
...
+class BuildlessSpec extends CommandSpec {
+    @Override
+    String getTempDirectoryPrefix() {
+        return "buildless"
+    }
+
+    @Unroll
+    void "test gradle buildless integration with #language"(Language language) {
+        when:
+        generateProject(language, BuildTool.GRADLE, ["buildless"])
```
timyates: Need to test BuildTool.GRADLE_KOTLIN as well

timyates (changes requested):

sgammon:
@timyates about the `BuildCacheConfiguration` -- we have two forms of integration with Gradle, a URL-only integration (which uses Gradle's standard `buildCache { ... }` block), and our plugin. Some companies don't want to use the plugin, and prefer the endpoint-only approach.

So the thinking there is that, eventually, if `micronaut-starter` gains parameterized `Feature`s, a user could pick between the plugin and the non-plugin approaches.

Then again, you're right that none of that functionality exists, nor planned to exist (to my knowledge), and could be considered irrelevant. I had stubbed it with a `static final` set to `false`, so the plugin is always used, anyway, so I figure I'll just remove the `BuildCacheConfiguration` and related code unless you object.

starter-core/src/main/java/io/micronaut/starter/feature/build/BuildCacheConfiguration.java
sgammon: Agreed, removed (easy to add in the future)

sgammon:
@timyates cc / @sdelamo I believe I've addressed all comments so far. I chose to leave out Maven integration for now, I can submit a followup PR for that. I implemented `GradleSpecificFeature` to cover the compatibility check with Maven.

sgammon:
By the way, if you guys would like a key to test with, let me know. :) Micronaut people can have Buildless for free, of course!

timyates:
> so I figure I'll just remove the `BuildCacheConfiguration` and related code unless you object.

I agree 👍

timyates:
Naming: Our community features usually have the name: "X-Y" where `X` is the name of the providing company, and "Y" is the feature name (ie: `agorapulse-micronaut-permissions` or `camunda-platform7`) but I'm not sure if that pattern fits here 🤔 

@sdelamo before this is merged, we need to do something about community-features and a maxMajor version check 🤔 

Also, should this be on 4.2.x?

starter-core/src/main/java/io/micronaut/starter/build/gradle/GradlePlugin.java
```diff
@@ -46,6 +46,7 @@ public class GradlePlugin implements BuildPlugin {
     private final boolean requiresLookup;
     private final List<GradleRepository> pluginsManagementRepositories;
     private final Set<String> buildImports;
+    private final Set<String> settingsImports;
     private final int order;
 
     public GradlePlugin(@NonNull GradleFile gradleFile,
```
sdelamo, line 78: deprecate the old constructor and create a new constructor.

starter-core/src/main/java/io/micronaut/starter/feature/CommunityFeatureValidator.java
```diff
@@ -24,7 +24,7 @@
 
 @Singleton
 public class CommunityFeatureValidator implements FeatureValidator {
-    public static final boolean ENABLE_COMMUNITY_FEATURES = false;
+    public static final boolean ENABLE_COMMUNITY_FEATURES = true;
```
sdelamo: merge 4.2.x into this branch

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/settingsGradle.rocker.raw
```diff
@@ -4,7 +4,15 @@
 @import io.micronaut.starter.build.gradle.GradleDsl
 @import java.util.Collection;
 @args (Project project, GradleBuild gradleBuild, boolean hasMultiProjectFeature, Collection<String> moduleNames)
+
+@if(gradleBuild.getDsl() == GradleDsl.KOTLIN && !gradleBuild.getSettingsImports().isEmpty()) {
```
sdelamo: why are we not adding the imports for the Groovy DSL?

starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
```diff
@@ -0,0 +1,100 @@
...
+    public String getCategory() {
+        return Category.DEV_TOOLS;
+    }
+
+    @Override
+    public String getThirdPartyDocumentation() {
+        return "https://docs.less.build/";
+    }
+
+    protected GradlePlugin gradlePlugin() {
```
sdelamo: I think all of these method can be extracted into a constant 

`private static final GradlePlugin BUILDLESS_GRADLE_PLUGIN`

starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
```diff
@@ -0,0 +1,100 @@
...
+import io.micronaut.starter.feature.Category;
+import io.micronaut.starter.feature.CommunityFeature;
+import io.micronaut.starter.feature.GradleSpecificFeature;
+import io.micronaut.starter.feature.build.gradle.templates.buildlessGradlePlugin;
+import io.micronaut.starter.template.RockerWritable;
+import jakarta.inject.Singleton;
+
+@Singleton
+public class Buildless implements CommunityFeature, GradleSpecificFeature {
+    public static final String NAME = "buildless";
```
sdelamo, line 31: make them all private unless use somehere

test-features/src/test/groovy/io/micronaut/starter/core/test/feature/buildless/BuildlessSpec.groovy
```diff
@@ -0,0 +1,40 @@
...
+import spock.lang.Unroll
+
+class BuildlessSpec extends CommandSpec {
+    @Override
+    String getTempDirectoryPrefix() {
+        return "buildless"
+    }
+
+    @Unroll
+    void "test gradle buildless integration with #language"(Language language) {
```
sdelamo: use a single test: 

```java
 @Unroll
    void "test #buildTool buildless integration with #language"(BuildTool buildTool, Language language) {
        when:
        generateProject(language, buildTool, ["buildless"])
        BuildResult result = executeGradle("compileJava")

        then:
        result?.output?.contains("BUILD SUCCESS")

        where:
       [buildTool, language] << [BuildTool.gradleValues(), Language.values()].combinations()
    }
```

starter-core/src/main/java/io/micronaut/starter/feature/build/BuildCacheConfiguration.java
sdelamo: can we delete this clasS?

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/build/BuildCacheConfiguration.java
sgammon: yes, sorry, it was meant to be deleted.

starter-core/src/main/java/io/micronaut/starter/build/gradle/GradlePlugin.java
sgammon: done

starter-core/src/main/java/io/micronaut/starter/feature/CommunityFeatureValidator.java
sgammon: rebased

starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/templates/settingsGradle.rocker.raw
sgammon: our plugin only needs it in the kotlin dsl -- good catch, not sure why i thought this was universal. i've fixed it in the latest push.

starter-core/src/main/java/io/micronaut/starter/feature/buildless/Buildless.java
sgammon: causes an exception with Rocker about templates only being one-shot renderable; i've restored it to non-static unless you have advice in this area

sgammon (commented):
looks clean from my end, i think i have addressed your comments, thank you for the review @timyates / @sdelamo.

this is now rebased on `4.2.x`, please let me know if any other changes are needed 👍🏻

starter-core/src/main/java/io/micronaut/starter/build/gradle/GradlePlugin.java
```diff
@@ -46,8 +46,10 @@ public class GradlePlugin implements BuildPlugin {
     private final boolean requiresLookup;
     private final List<GradleRepository> pluginsManagementRepositories;
     private final Set<String> buildImports;
+    private final Set<String> settingsImports;
     private final int order;
 
+    @Deprecated
```
timyates:
```suggestion
    @Deprecated(since = "4.2.0", forRemoval = true)
```

timyates: And can we change the deprecated constructor to just call:

```java
        this(
                gradleFile,
                id,
                version,
                artifactId,
                extension,
                settingsExtension,
                pluginsManagementRepositories,
                requiresLookup,
                order,
                buildImports,
                Collections.emptySet()
        );
```

As it reduces our surface area for future mistakes 😉

starter-core/src/test/groovy/io/micronaut/starter/feature/buildless/BuildlessSpec.groovy
```diff
@@ -0,0 +1,68 @@
...
+        expect:
+        for (applicationType in ApplicationType.values()) {
+            assert feature.supports(applicationType)
+        }
+
+        feature.category == Category.DEV_TOOLS
+        feature.name == Buildless.NAME
+    }
+
+    void "buildless configured correctly for #buildTool"() {
```
timyates, line 52:
```suggestion
    void "feature buildless is unsupported with maven and #language"(Language language) {
        when:
        generate(ApplicationType.DEFAULT, new Options(language, BuildTool.MAVEN), [Buildless.NAME])

        then:
        def e = thrown(IllegalArgumentException)
        e.message == 'Feature only supported by Gradle'

        where:
        language << Language.values()
    }

    void "buildless configured correctly for #buildTool"() {
```

timyates (changes requested):
So close! 😁 

Couple of suggestions, and you'll need to fix the `FeatureControllerSpec#test community features` test.  We can do:

```diff
diff --git a/starter-api/src/test/groovy/io/micronaut/starter/api/FeatureControllerSpec.groovy b/starter-api/src/test/groovy/io/micronaut/starter/api/FeatureControllerSpec.groovy
index 1264b70b12..c3dceb5af7 100644
--- a/starter-api/src/test/groovy/io/micronaut/starter/api/FeatureControllerSpec.groovy
+++ b/starter-api/src/test/groovy/io/micronaut/starter/api/FeatureControllerSpec.groovy
@@ -27,14 +27,15 @@ class FeatureControllerSpec extends Specification {

     void "test community features"() {
         given:
-        List<String> expected = [
+        Set<String> expected = [
+                'buildless',
                 'jobrunr-jobrunr'
         ]
         when:
         List<FeatureDTO> communityFeatures = client.features(ApplicationType.DEFAULT, RequestInfo.LOCAL).features.findAll { it.community }

         then:
-        communityFeatures.name.sort() == expected.sort()
+        communityFeatures.name =~ expected
     }
```

To add `buildless` and do set comparison in Spock

We also need to think about naming.

We prefix community features with the company name (see `jobrunr-jobrunr` and `camunda-zeebe`)...  I think we should keep this pattern here 🤔 @sdelamo ?

sgammon:
@timyates Okay, no problem! My power is out but as soon as I'm back to my desk I should be able to fix.

Re/naming, Buildless is an open source tool and is the name of the company and product, technically. If there's any way we can keep it "buildless," that would be our preference, but if needed we can change to "elide-buildless", as there is an affiliated company called Elide. If there is a strong need for this let me know and I'll make this change as well.

sgammon:
(We own the domain less.build, so everything really is under that name. Our Maven and Gradle coordinates are grouped under build.less 😂, a bit of a hack.)

sgammon:
@timyates / @sdelamo next round of review comments done :)

starter-core/src/test/groovy/io/micronaut/starter/feature/buildless/BuildlessSpec.groovy
timyates: Did we not add the test for maven?

sgammon: @timyates I adopted 'GradleFeature' to make this Gradle-only for now, as our Maven integration is in beta. I intend to return with a new PR for Maven after the U.S. holiday.

sgammon:
all tests passing on my end, @timyates I've adopted your latest updates -- the Maven changes will follow later, for now, the feature adopts `GradleFeature`.

timyates (approved):
Nice!

Tried signing up to the beta at https://less.build/private-beta to give it a final functional test, but the form is failing atm 😢 

image

sgammon:
@timyates Weird! We will take a look. I think I actually got your signups, we'll get you a key!

timyates:
@sgammon Hiya!  Cool, got my key 🙏

So if I create an app with the kotlin dsl, and another with the groovy dsl via:

```
./gradlew micronaut-cli:run --args="create-app --features=buildless com.example.demo"
```

and

```
./gradlew micronaut-cli:run --args="create-app --build=gradle --features=buildless com.example.demo2"
```

Then, add `BUILDLESS_APIKEY` as an env var, and run `./gradlew build --scan` in both `starter-cli/demo` (kotlin dsl) and `starter-cli/demo2` (groovy dsl)

The remote cache is configured in the Kotlin dsl build:

https://scans.gradle.com/s/kra64ipgi26fc/performance/build-cache#remote-cache

But it seems to be disabled for the Groovy DSL build:

https://scans.gradle.com/s/vtfc36zeokmhq/performance/build-cache#remote-cache

Any ideas?

sgammon:
@timyates no, no idea at all! But we will take a look and I will get back to you shortly. 😄 

Thanks for giving this a thorough test.

sgammon:
@timyates Okay, thanks for reporting that -- it helped us figure out a number of issues in the plugin. We've solved those with a few new releases, culminating in ~~`1.0.0-beta5`~~ `1.0.0-beta6` (which I've updated here).

This work is sort of colliding in time with our new [Buildless Agent](https://docs.less.build/docs/cli), which runs via our [CLI](https://github.com/buildless/cli/releases). There was already work in motion to enable the Gradle plugin to detect and use the Agent if it is running locally.

All the above has now landed, resulting in:

- No need for an import in the Kotlin DSL (it's been removed)
- Working Buildless via the Groovy DSL
- Buildless Agent support
- No change to plugin behavior otherwise (when using the cloud service, via `BUILDLESS_APIKEY`)

I've verified that Groovy works in my environment but it would be greatly appreciated if you could confirm.

P.S. The Buildless Agent, once fully released, will be free forever, as in beer, for unlimited users/builds/CPUs/everything. Entirely optional, but we'd love you to give it a try. The Agent is usable with or without Cloud backhaul. If you _do_ try the agent, it should pick up your API key from `BUILDLESS_APIKEY` and automatically enable Cloud services.


<details>
<summary>Edit: This part is now irrelevant</summary>
There is at least one known issue, but it involves multi-module project structures and some other stuff which is unrelated here.
</details>

sgammon:
Ah, hold up, I may have spoken too soon on the Groovy front. Stay tuned.

sgammon:
Okay, sorry, @timyates, this should actually be done now. Groovy is working as expected and there is actually no need for (a) an import, or (b) the `buildless { }` call anymore.

The plugin will activate if the following conditions are met:

- The plugin is added to the Gradle build (obviously)
- The agent is running, detected via a known place on disk, in which case the agent is used, OR
- An API key is detected, in which case the Cloud service is used

If the first or both of the latter conditions fail, the plugin is inert, aside from fixing up other things (withholding larger Micronaut-related outputs like `shadowJar` automatically is a feature that landed in `beta6`).

Please let me know if you find anything amiss. Everything looks good with local testing; I have one Spring-related test which is failing, but I haven't seen that fail anywhere else, including CI, so I'm ignoring it for now.

sgammon:
friendly bump for re-review @sdelamo :) ?

timyates (approved):

sgammon:
@sdelamo I think these changes are ready to go! Sorry for the multiple rounds of review

sdelamo (approved):

sgammon:
@sdelamo, @timyates, thank you both for your review and guidance here 😄

PR #2182, opened 2023-11-15
Adds Micronaut Community Feature

sdelamo:
This pull request adds an extra marker interface MicronautCommunityFeature.

- MicronautCommunityFeature marks a feature that contributes a Micronaut library. Those features specify the major version of Micronaut they were built with.

- CommunityFeature marks a feature that does not contribute a Micronaut Library. It may contribute a third-party Java library or build feature. For example, the build.less feature

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/MicronautCommunityFeature.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/MicronautCommunityFeature.java
@@ -19,7 +19,7 @@ import io.micronaut.starter.options.MicronautVersion;
 import io.micronaut.starter.util.VersionInfo;
 
 /**
- * Markert interface for {@link CommunityFeature} which expose a Micronaut Library.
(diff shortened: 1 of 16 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/MicronautCommunityFeature.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.starter.feature;
+
+import io.micronaut.starter.options.MicronautVersion;
+import io.micronaut.starter.util.VersionInfo;
+
+/**
+ * Markert interface for {@link CommunityFeature} which expose a Micronaut Library.
```
timyates:
```suggestion
 * Marker interface for {@link CommunityFeature} which expose a Micronaut Library.
```

timyates (approved):

PR #2192, opened 2023-11-21
Convert JdkVersion from enum to class

burtbeckwith:
Changing from an enum allows adding new instances via the constructor (e.g. when using starter as a library) without recompiling.


sdelamo (changes requested):
Please, change the PR base to 4.3.x

burtbeckwith:
@sdelamo ok to merge?

sdelamo (changes requested):
Can we do this change in a backwards compatible way so that the JSON serialization remains? See test: https://github.com/micronaut-projects/micronaut-starter/pull/2192/commits/7faa53597a0eaa8aff35da3fdd2bddcd80a62d71

sdelamo:
> Can we do this change in a backwards compatible way so that the JSON serialization remains? See test: [7faa535](https://github.com/micronaut-projects/micronaut-starter/commit/7faa53597a0eaa8aff35da3fdd2bddcd80a62d71)

I tried annotating JdkVersion::name with `@JsonValue` but that breaks `SelectOptionsControllerTest` test.

sdelamo:
I created a issue for a bug in serde which affects this PR https://github.com/micronaut-projects/micronaut-serialization/pull/709

PR #2201, opened 2023-11-29
Telegram ChatBot support in Starter

timyates:
This adds features for Telegram chatbots based on

- Azure
- Lambdas
- Gcp
- Http

When used, the generated README explains how to set up and deploy the bot.

As part of the feature, an example `about` command is added as well as a
command that catches any input that is unrecognised.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/aws/Cdk.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/aws/Cdk.java
@@ -55,7 +55,6 @@ import io.micronaut.starter.feature.build.gradle.templates.useJunitPlatform;
 import io.micronaut.starter.feature.build.maven.templates.execMavenPlugin;
 import io.micronaut.starter.feature.build.maven.templates.genericPom;
 import io.micronaut.starter.feature.build.maven.templates.mavenCompilerPlugin;
-import io.micronaut.starter.feature.chatbots.telegram.TelegramAwsChatBot;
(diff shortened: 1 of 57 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/aws/Cdk.java
```diff
@@ -148,7 +149,9 @@ public void apply(GeneratorContext generatorContext) {
         generatorContext.addTemplate("cdk-main", new RockerTemplate(INFRA_MODULE, "src/main/java/{packagePath}/" + MAIN_CLASS_NAME + ".java",
                 cdkmain.template(generatorContext.getProject())));
 
-        String handler = HandlerClassFeature.resolveHandler(generatorContext);
+        String handler = generatorContext.isFeaturePresent(TelegramAwsChatBot.class) ?
```
sdelamo: `TelgramAwsChatBot` should be a `HandleClassFeature`

sdelamo (changes requested):
I suggested some changes https://github.com/micronaut-projects/micronaut-starter/pull/2226

starter-core/src/main/java/io/micronaut/starter/feature/aws/Cdk.java
timyates: This is done in your PR (now merged), yeah?

sdelamo (approved):

PR #2223, opened 2023-12-06
Mergeup 4.2.x

timyates:

PR #2233, opened 2023-12-08
Fix flakiness in SwaggerUISpec

timyates:
Yaml isn't in a guaranteed order so the comparison was flaky.

Replaced by parsing it into maps and comparing those maps with Groovy


timyates:
This is cherry-picked from the 4.3.x branch where it was discovered https://github.com/micronaut-projects/micronaut-starter/pull/2223

sdelamo (approved):

PR #2244, opened 2023-12-11
test: JdkVersion serde test

sdelamo:
see: #2192


timyates:
Why introspected in starter-api, but Serdeable in starter-web-netty?

Is this to check they are both working?

timyates (approved):

sdelamo:
> Why introspected in starter-api, but Serdeable in starter-web-netty?
> 
> Is this to check they are both working?

it is nice side effect.

PR #2245, opened 2023-12-11
Add `create-chatbot` command to CLI

### LINKED ISSUE #2229: create `create-chatbot-telegram` CLI command
> $ create-chatbot-telegram
> 
> How do you want to deploy it?
> 
> *1) AWS Lambda Function
> 2) Google Cloud Function
> 3) Azure Function

ApplicationType will be `FUNCTION` automatically. 

For Lambda the `aws-lambda-function-url` feature is added automatically.

For Lambda the following two questions will be asked: 

> 
> How do you want to deploy?. (enter for Java runtime)
> *1) Fat JAR running in the managed Java runtime
>  2) Native executable built with GraalVM Native Image running in a custom runtime. Faster cold start
> > 
> 
> Choose your Lambda Architecture. (enter for x86)
>  1) arm
> *2) x86
> > 

> 
> Do you want to generate infrastructure as code with CDK? (enter for yes)
> *1) YES
>  2) NO

for every cloud the follow questions will be asked ask well. 

> Choose your preferred language. (enter for default)
> *1) Java
>  2) Groovy
>  3) Kotlin
> > 
> 
> Choose your preferred test framework. (enter for default)
> *1) JUnit
>  2) Spock
>  3) Kotest
> > 
> 
> Choose your preferred build tool. (enter for default)
>  1) Gradle (Groovy)
> *2) Gradle (Kotlin)
>  3) Maven
> > 
> 
> Choose the target JDK. (enter for default)
> *1) 17
> > 
> 

> Enter any features to apply. Use tab for autocomplete and separate by a space.

See `create-aws-lambda` command. There is a lot of common code. 

[Azure functions support only 17. ](https://learn.microsoft.com/en-us/azure/azure-functions/functions-versions?tabs=isolated-process%2Cv4&pivots=programming-language-java)

[Google Cloud function support 21 but only in 2nd gen preview only](https://cloud.google.com/functions/docs/concepts/java-runtime)

AWS Lambda supports 21

timyates:
This adds a new command `create-chatbot` to the `mn` cli tool.

I've extracted some common/repeated code up to the top level class.

I decided to not add `telegram` to the name at the moment, as more chatbots are coming and we should be able to use the same command....

```
./gradlew :micronaut-cli:build    
 java -jar starter-cli/build/libs/micronaut-cli-4.3.0-SNAPSHOT-all.jar create-chatbot
```

### DIFF
--- a/starter-cli/src/main/java/io/micronaut/starter/cli/MicronautStarter.java
+++ b/starter-cli/src/main/java/io/micronaut/starter/cli/MicronautStarter.java
@@ -26,6 +26,7 @@ import io.micronaut.starter.cli.command.BuildToolConverter;
 import io.micronaut.starter.cli.command.CodeGenCommand;
 import io.micronaut.starter.cli.command.CreateAppCommand;
 import io.micronaut.starter.cli.command.CreateBuilderCommand;
+import io.micronaut.starter.cli.command.CreateChatBotBuilderCommand;
 import io.micronaut.starter.cli.command.CreateCliCommand;
 import io.micronaut.starter.cli.command.CreateFunctionCommand;
 import io.micronaut.starter.cli.command.CreateGrpcCommand;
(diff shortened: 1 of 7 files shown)


sdelamo (commented):
We should tell the user they are creating a Telegram BOT. 

Even if we give them a prompt with a single answer. I think we should ask them something like this in the beginning: 

> What kind of ChatBot do you want to develop?
*1) Telegram ChatBot

Thoughts?

timyates:
I was in 2 minds whilst doing this...  But I think you're right, as it serves as documentation as well as leaving a spot for future types

Pushed a change so the first question is:

```
Choose your ChatBot type. (enter for Telegram)
*1) Telegram - A chat bot for the telegram messaging platform
>
```

sdelamo (approved):

PR #2263, opened 2024-01-04
Add Basecamp ChatBot features

timyates:
And update the create-chatbot CLI command to support them


timyates:
Going to run some functional tests on this pre final release of chatbots

starter-cli/build.gradle
```diff
@@ -77,6 +77,10 @@ startScripts {
     applicationName = 'mn'
 }
 
+run {
+    standardInput = System.in
```
sdelamo, line 81: why is this needed?

starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateChatBotBuilderCommand.java
```diff
@@ -114,21 +115,36 @@ enum ChatBotType {
...
-        private final String feature;
+        private final String basecampFeature;
+        private final String telegramFeature;
         private final ApplicationType applicationType;
 
-        ChatBotDeployment(String title, String description, String feature, ApplicationType applicationType) {
+        ChatBotDeployment(
+                String title,
+                String description,
+                String basecampFeature,
```
sdelamo: can we use a list or a `Map<ChatBotType, String>` so that each time we add a new chatbot we don't have to modify the constructor?

test-features/src/test/groovy/io/micronaut/starter/core/test/feature/chatbots/basecamp/BasecampGcpChatBotFunctionSpec.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+    void "#testFramework test #feature feature in #language with #buildTool"(BuildTool buildTool, Language language, TestFramework testFramework) {
+        when:
+        generateProject(language, buildTool, [feature], ApplicationType.FUNCTION, testFramework)
+
+        then:
+        String result = executeBuild(buildTool, "test")
+
+        then:
+        println result
```
sdelamo:
```suggestion
```

sdelamo (changes requested):

starter-cli/build.gradle
timyates: Without this, you can't run `./gradlew :m-cli:run --args='create-chatbot'`

starter-cli/src/main/java/io/micronaut/starter/cli/command/CreateChatBotBuilderCommand.java
timyates: Switched to accept a function, this means we can use the new Java 17 switch construct and make sure we don't miss any

https://github.com/micronaut-projects/micronaut-starter/pull/2263/commits/a4a7d886109e6aa0bf1d2920cef70f153751c2f3

What do you think?

sdelamo (approved):

PR #2291, opened 2024-02-01
lambda default to Jackson Databind instead of serde

sdelamo:
it is slightly faster to use Jackson databind

CleanShot 2024-02-01 at 10 42 25@2x

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/jaxrs/JaxRsSecurity.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/jaxrs/JaxRsSecurity.java
@@ -17,7 +17,6 @@ package io.micronaut.starter.feature.jaxrs;
 
 import io.micronaut.starter.application.ApplicationType;
 import io.micronaut.starter.application.generator.GeneratorContext;
-import io.micronaut.starter.build.dependencies.Dependency;
 import io.micronaut.starter.build.dependencies.MicronautDependencyUtils;
 import io.micronaut.starter.feature.Category;
 import io.micronaut.starter.feature.Feature;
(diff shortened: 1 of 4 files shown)


graemerocher:
personally don't think we should make this change. Instead we should profile startup of serde and see how it can be improved. //cc @dstepanov

dstepanov:
I'm planning to improve the startup time of serde

sdelamo:
> personally don't think we should make this change. Instead we should profile startup of serde and see how it can be improved. //cc @dstepanov

Once we have a faster startup in serde we should default to it. But for Serverless, we should default to the fastest settings as many users would evaluate whether to use Micronaut based on cold startup performance.

timyates (approved):

dstepanov:
I did some benchmarking, and I cannot see the slower performance of serde:

```
Benchmark                                                  (stack)  Mode  Cnt  Score   Error  Units
StartupBenchmark.fullStartupAndGetMapper             SERDE_JACKSON    ss   10  0.459 ± 0.022   s/op
StartupBenchmark.fullStartupAndGetMapper          JACKSON_DATABIND    ss   10  0.566 ± 0.060   s/op
StartupBenchmark.fullStartupAndGetMapperReadBean     SERDE_JACKSON    ss   10  0.476 ± 0.047   s/op
StartupBenchmark.fullStartupAndGetMapperReadBean  JACKSON_DATABIND    ss   10  0.599 ± 0.013   s/op
StartupBenchmark.getMapper                           SERDE_JACKSON    ss   10  0.044 ± 0.004   s/op
StartupBenchmark.getMapper                        JACKSON_DATABIND    ss   10  0.143 ± 0.022   s/op
StartupBenchmark.getMapperReadBean                   SERDE_JACKSON    ss   10  0.088 ± 0.013   s/op
StartupBenchmark.getMapperReadBean                JACKSON_DATABIND    ss   10  0.195 ± 0.012   s/op
```

dstepanov:
What version did you use in the lambda test?

PR #2320, opened 2024-02-09
Add unit test

sdelamo:
I don't know if this adds any value or it is better to leave it without a test.


starter-cli/src/main/java/io/micronaut/starter/cli/command/BuilderCommand.java
```diff
@@ -105,6 +103,12 @@ public Integer call() throws Exception {
         return 0;
     }
 
+    public Set<String> getFeatures(GenerateOptions options, Supplier<List<String>> featureSupplier) {
```
timyates, line 106: This doesn't need to be a supplier I don't think 🤔

timyates (approved):

PR #2346, opened 2024-02-14
disable oracle for hibernate reactive

sdelamo:
This reverts commit 42ce7f799ba3d938c405e947b6af48d127a967a8.


radovanradic:
Looks like related to the revert for ojdbc version done in micronaut-sql. In micronaut-sql hibernate-reactive test for oracle picks up 23.x version through vertx oracle client 4.5.3 and has no issues. Then I tried to force version 21.x and got the same error. It's been also talked about in [stackoverflow](https://stackoverflow.com/a/77732758).
So, it seems vertx expects 23.x version and we changed back to 21.x. So maybe should downgrade vertx oracle version too? Oracle 23.x dependency was introduced in vertx 4.5.0

sdelamo:
The test passes with Gradle it fails for Maven. In the SQL module we test with Gradle as well.

sdelamo:
This PR is a workaround. It generates projects that use `2.3.0.23.09` for Hibernate Reactive and Oracle. 

However, I think the correct thing would be to rollback this PR https://github.com/micronaut-projects/micronaut-sql/pull/1268 

In that PR, we went back two major versions of `ojdbc` to support GraalVM Native image. However, that version breaks the support for Hibernate Reactive and Oracle. I think it would be best to back to the modern version of ojdbc and document it is currently broken for GraalVM Native image.

sdelamo:
I have disabled Oracle for Hibernate reactive. I have refactored the tests so that it is easy to enable it again once it is supported.

starter-core/src/test/groovy/io/micronaut/starter/feature/database/DataHibernateReactiveSpec.groovy
```diff
@@ -344,4 +350,33 @@ class DataHibernateReactiveSpec extends BaseHibernateReactiveSpec {
...
+        scenarios.stream()
+                .filter(ts -> ts.db != Oracle.NAME || Oracle.COMPATIBLE_WITH_HIBERNATE_REACTIVE)
+                .toList()
+    }
+
+    private List<TestScenario> filter(List<TestScenario> scenarios) {
+        scenarios.stream()
+                .filter(ts -> ts.db != Oracle.NAME || Oracle.COMPATIBLE_WITH_HIBERNATE_REACTIVE)
+                .toList()
+    }
```
timyates, line 381: `list.findAll`, Java boy! 😉

timyates (approved):

PR #2347, opened 2024-02-14
apply ojdbc11 not ojdbc8

sdelamo:
I assume we should apply ojdbc11 instead of ojdbc8

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/database/Oracle.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/database/Oracle.java
@@ -31,9 +31,17 @@ public class Oracle extends DatabaseDriverFeature {
 
     public static final String NAME = "oracle";
     public static final String VERTX_ORACLE_CLIENT = "vertx-oracle-client";
+
+    @Deprecated(forRemoval = true)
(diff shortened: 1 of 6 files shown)


radovanradic (approved):

starter-core/src/main/java/io/micronaut/starter/feature/database/Oracle.java
```diff
@@ -31,12 +31,20 @@ public class Oracle extends DatabaseDriverFeature {
 
     public static final String NAME = "oracle";
     public static final String VERTX_ORACLE_CLIENT = "vertx-oracle-client";
+
+    @Deprecated(forRemoval = true)
```
alvarosanchez, line 35: IMO it doesn't make sense to deprecate this. It should be removed. I know you did it for strict SemVer compatibility, but honestly, this particular case should be removed

alvarosanchez (approved):

starter-core/src/main/java/io/micronaut/starter/feature/database/Oracle.java
sdelamo: I am trying to be conservative about unnecessary breaking changes to make the GCN Launcher integration easier.

PR #2357, opened 2024-02-17
Fixes in ADoc

altro3:


.github/workflows/release.yml
```diff
@@ -4,7 +4,7 @@ on:
     types: [published]
 jobs:
   build:
-    name: Release artifacts to Maven Central & SDKMAN, and publish documentation
+    name: Release artifacts to Maven Central & SDKMAN!, and publish documentation
```
sdelamo, line 7: please revert the addition of !. We don't need to add ! to the workflow names.

.github/workflows/release.yml
```diff
@@ -261,7 +261,7 @@ jobs:
           release_version=${GITHUB_REF:11}
           sed -i "s/^projectVersion.*$/projectVersion\=${release_version}/" gradle.properties
           echo "release_version=${release_version}" >> $GITHUB_OUTPUT
-      - name: SDKMAN release
+      - name: SDKMAN! release
```
sdelamo, line 264: please revert the addition of !. We don't need to add ! to the workflow names.

sdelamo (approved):

PR #2378, opened 2024-02-26
major: Switch to security-processor for Micronaut 5.0.0

timyates:
In https://github.com/micronaut-projects/micronaut-security/pull/1611 for Micronaut 5 we moved the JSR 250 annotation mappers to a new module.

This PR updates starter to use this module instead of Micronaut-annotations.

It is a breaking change as it renames the hidden SecurityAnnotations feature to SecurityProcessor.


timyates:
Switched to a draft as this won't work until securty is released and we have a 5.0.0-SNAPSHOT of platform

sdelamo (approved):

PR #2395, opened 2024-03-08
Azure Key Vault needs micronaut discovery client dependency.

wetted:
see https://github.com/micronaut-projects/micronaut-azure/issues/570


sdelamo (changes requested):
This should apply the `DiscoveryClient`feature

PR #2419, opened 2024-03-25
Remove .set() usage from Kotlin DSL

### LINKED ISSUE #2417: Remove set() from Kotlin DSL for Gradle
It was noticed that our Kotlin DSL scripts still use the `.set(....)` syntax.
Since Gradle 8.2, we can use the `=` operator, that will make our scripts nicer.

ie: the aot configuration

```
        convertYamlToJava.set(false)
```

can become

```
        convertYamlToJava = false
```

timyates:
Since Gradle 8.2 it should not be required

Closes #2417


timyates:
I don't think the failures are related to this PR, but I'll investigate a fix...

sdelamo (approved):

timyates:
Yay!

PR #2422, opened 2024-03-26
Switch to use oracle-free instead of oracle-xe

timyates:
In 2.4.0 of TestResources a new Oracle provider was added and the old one deprecated.

For some reason, the old one is failing in our tests for 4.4.0 with

```
Test resources doesn't support resolving expression 'datasources.default.password'
```

https://ge.micronaut.io/s/igklqice44n7w/tests/overview?outcome=FAILED

The fix seems to be to switch to use `KnownModules.JDBC_ORACLE_FREE` instead...

Will people start seeing this failure if they don't change the module in their build when they upgrade?


melix (approved):
The new version changes the inference module to use Oracle Free instead of Oracle XE. However it is indeed curious that the upgrade would break existing builds, as if the old module wasn't used?

sdelamo (approved):

PR #2474, opened 2024-04-26
Generate an ARM OSX binary and publish it to SDKMan and Homebrew

### LINKED ISSUE #2473: Generate native ARM binaries for OSX
Currently we ship an x86 binary for both the arm and x86 channels of SDKMan, then let Rosetta sort it out on the client machine.

[We tried previously to generate an ARM binary](https://github.com/micronaut-projects/micronaut-starter/issues/2107) but at the time there were no ARM runners to build on.

Yesterday, Github switched the `macos-latest` runner across to OSX 14, and it only runs on ARM hardware.

So we should use `macos-latest` to generate the ARM binary.

And `macos-13` to generate the x86 binary.

And then push the correct binary to the correct SDKMan channel

timyates:
The `macos-latest` github runner is now an ARM machine.

To build the amd64 image, we now need to use `macos-13`.

We can turn this breaking change into a positive by using `macos-latest` to generate an ARM binary of the CLI tool, and `macos-13` to generate an x86 binary.

These can then be published to the ARM and X86 channels of SDKMan and Homebrew

Homebrew cask change is here https://github.com/micronaut-projects/homebrew-tap/pull/7

Closes #2473


sdelamo (approved):

graemerocher (approved):

sdelamo:
@timyates resolve the conflict please.

PR #2483, opened 2024-05-08
Fix Maven/Groovy builds with zipkin and jaeger

timyates:
We had an api dependency on micronaut-core-processor which meant it was ending up on the runtime classpath.

This was fixed here

https://github.com/micronaut-projects/micronaut-tracing/issues/553#event-12652214705

However for Maven and Groovy this causes a failure as it no longer ends up getting pulled onto the provided configuration.

Once this fix is released, we can remove the skips from the jaeger and zipkin guides here:

https://github.com/micronaut-projects/micronaut-guides/pull/1459/commits/798040a62c910ec8ac19852e89c68f79aa4c12f0#diff-429b78dd62241a22ac3b663ca2d045441a9326121c3b6bbf19f5859d94c2e3d8

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/build/dependencies/MicronautDependencyUtils.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/dependencies/MicronautDependencyUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2022 original authors
+ * Copyright 2017-2024 original authors
  *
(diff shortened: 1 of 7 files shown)


sdelamo (approved):

sdelamo:
@timyates create a draft PR in guides reverting the skips

timyates:
https://github.com/micronaut-projects/micronaut-guides/pull/1461

PR #2512, opened 2024-06-13
Add React SSR feature

mikehearn:
Please note that this PR is against the base 5.0.x branch

Adds support for the recently added React SSR view renderer. The generated project uses NPM plugins for Gradle and Maven to set up an integrated frontend/backend project with WebPack driven by the build, so it's a little more complex than a normal template renderer feature.


graemerocher:
should target the 4.5.x branch, it can go in a patch release

mikehearn:
The React SSR feature itself isn't released yet though, right? That was merged into the 5.x branch

graemerocher:
it has been released https://micronaut-projects.github.io/micronaut-views/latest/guide/#react

It is in 5.4.x but that is the version number of `micronaut-views` module not micronaut itself

PR #2524, opened 2024-06-19
Add React SSR feature

mikehearn:
Re-opened on 4.5.x

This PR adds support for creating a sample project that uses React server side rendering (SSR).

### DIFF
--- a/buildSrc/src/main/java/io/micronaut/starter/coordinates/CoordinatesSourceGenerator.java
+++ b/buildSrc/src/main/java/io/micronaut/starter/coordinates/CoordinatesSourceGenerator.java
@@ -98,6 +98,7 @@ public abstract class CoordinatesSourceGenerator extends DefaultTask {
         getFileOperations().delete(outputDirectory);
         Files.createDirectories(packageDirectory);
         try (PrintWriter writer = new PrintWriter(new FileWriter(new File(packageDirectory.toFile(), "StarterCoordinates.java")))) {
+            writer.println("// WARNING: Generated from gradle/templates.versions.toml - DO NOT EDIT.");
             writer.println("package " + packageName + ";");
             writer.println();
             writer.println("import java.util.HashMap;");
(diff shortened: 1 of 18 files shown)


graemerocher:
@melix could you look at the generated Gradle config, not sure if it represents best practise

melix:
I will. At first glance there are improvements which can be made to avoid `dependsOn` but it depends on (ah ah!) the plugins. I need to try this to figure out.

mikehearn:
OK. What's wrong with dependsOn? Do you have a link to the current Gradle best practices?

melix:
I wrote a blog post about this a while back : https://melix.github.io/blog/2021/10/gradle-quickie-dependson.html

In general you should have a task which produces an output, then the tasks which consume this output must define it as an input : `dependsOn` should be limited to lifecycle tasks. However, as I said, sometimes because the plugins we use are not using say the `Provider` API, it's not always possible.

mikehearn:
I think in this case it wouldn't make a difference. The custom `build{Client,Server}JS` tasks that produce files do declare inputs and outputs. The other tasks are just for grouping and CLI usage purposes, they don't consume the inputs of the tasks they depend on. The NPM install task, I don't know what outputs it has indeed. Probably `processResources` should be depending on the JS tasks via their outputs but I don't know how to do that either without something overkill like creating a whole custom plugin. Gave up trying to understand Gradle years ago, honestly ...

melix:
`processResources` is precisely the task I was referring to. Lifecycle tasks are ok. The thing if you don't wire properly is that builds work accidentally. For Maven it's often not a problem because folks run `clean install` all the time so they often don't notice. Anyway, I'll take a look soon.

mikehearn:
This PR depends on https://github.com/micronaut-projects/micronaut-views/pull/833 as otherwise the config needed is excessively complicated.

mikehearn:
@melix Thanks for your review, I've force pushed a new version of the commit that uses `builtBy`. Once there's a new Micronaut Views release I'll update the versions in this PR to make sure it's being used and then we can merge it.

melix (approved):

graemerocher:
can you correct the check style failures?

PR #2541, opened 2024-08-05
Some fixes to the default ReactJS template.

mikehearn:
1. Bump the graaljs library version. This avoids a confusing configuration error message that can occur with the default setup.
2. Don't use a POM dependency on GraalJS for Gradle, this is incorrect and can yield invalid "version mismatch" errors depending on your JVM.
3. Configure Gradle to download GraalVM automatically.

Together these should improve the rate at which projects work out of the box.


graemerocher (approved):

mikehearn:
Thanks. Can someone merge?

PR #2548, opened 2024-08-23
Fix build

sdelamo:


starter-web-servlet/src/test/groovy/io/micronaut/starter/servlet/OpenAPISpec.groovy
```diff
@@ -18,19 +19,42 @@ class OpenAPISpec extends Specification {
     @Inject
     HttpClient client
 
-    void "test swagger-ui"() {
+    @Ignore
+    void "test swagger-ui links"() {
```
sdelamo, line 23: @altro3 has the html output changed for swagger-ui and rapid? These tests which verified the links fail now.

starter-api/src/main/java/io/micronaut/starter/api/diff/DiffController.java
```diff
@@ -179,49 +177,45 @@ public Publisher<String> diffApp(
...
-                emitter.next("");
-                featureDiffer.produceDiff(
-                        projectGenerator,
-                        generatorContext,
-                        new ConsoleOutput() {
-                            @Override
-                            public void out(String message) {
-                                emitter.next(message + LINE_SEPARATOR);
-                            }
+    private String diff(ProjectGenerator projectGenerator, GeneratorContext generatorContext) {
```
sdelamo, line 183: @yawkat @dstepanov have we changed how `Publisher<String>` is managed? I had to change this. The previous code was timing out in the test.

starter-web-servlet/src/test/groovy/io/micronaut/starter/servlet/OpenAPISpec.groovy
altro3: Yes, links were changed. Chanes here: https://github.com/micronaut-projects/micronaut-openapi/pull/1626

Currently, links to scripts are not explicitly specified, and scripts are loaded dynamically. The code in the test must be changed

starter-api/src/main/java/io/micronaut/starter/api/diff/DiffController.java
yawkat: could be https://github.com/micronaut-projects/micronaut-core/pull/11107

PR #2550, opened 2024-08-23
Add support for HTTP POJA feature

andriy-dmytruk:
Add a feature based on this HTTP POJA: https://github.com/micronaut-projects/micronaut-servlet/pull/773
Also HTTP POJA runtime in micronaut-gradle-plugin: https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/1024

The module was not yet published, so this cannot be merged yet.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/server/HttpPoja.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/server/HttpPoja.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/server/HttpPoja.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public String getMicronautDocumentation() {
+        return "https://micronaut-projects.github.io/micronaut-servlet/latest/guide/index.html#http-poja";
+    }
+
+    @Override
+    public void doApply(GeneratorContext generatorContext) {
+        if (generatorContext.getBuildTool() == BuildTool.MAVEN) {
```
sdelamo, line 49: why are the dependencies only applied for Maven builds?

andriy-dmytruk: > why are the dependencies only applied for Maven builds?

I added a PR to the gradle plugin for another runtime: https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/1024

I think that makes sense since it is a similar server implementation to other runtimes

andriy-dmytruk:
@sdelamo Can we merge this?

graemerocher (approved):

PR #2554, opened 2024-08-24
handle HTML request correctly

sdelamo:
@yawkat @dstepanov there is a regression which broken the redirection from 

https://launch.micronaut.io to https://micronaut.io/launch

This PR fixes it. I think the fix is the correct way of writing this controller but I am not sure if we want to support the old behaviour.


dstepanov (approved):

yawkat:
Why does the old way not work anymore

dstepanov:
Probably https://github.com/micronaut-projects/micronaut-core/issues/11114#issuecomment-2308980717

yawkat:
I have nothing against this change but maybe it would be better to just wait for the core fix

PR #2604, opened 2024-09-09
add jte-kotlin if ktolin

sdelamo:
See: https://github.com/micronaut-projects/micronaut-views/issues/876

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/view/JTE.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/view/JTE.java
@@ -18,6 +18,7 @@ package io.micronaut.starter.feature.view;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.starter.application.generator.GeneratorContext;
 import io.micronaut.starter.build.dependencies.Coordinate;
+import io.micronaut.starter.build.dependencies.Dependency;
 import io.micronaut.starter.build.dependencies.MicronautDependencyUtils;
 import io.micronaut.starter.build.gradle.GradlePlugin;
 import io.micronaut.starter.build.gradle.GradleDsl;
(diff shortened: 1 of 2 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/view/JTE.java
```diff
@@ -80,6 +81,18 @@ public void apply(GeneratorContext generatorContext) {
         generatorContext.addTemplate("exampleJte", new RockerTemplate(JTE_SRC_DIR + "/example.jte", exampleJTE.template()));
     }
 
+    protected void addDependencies(GeneratorContext generatorContext) {
+        generatorContext.addDependency(MicronautDependencyUtils.viewsDependency()
+                .artifactId(ARTIFACT_ID_MICRONAUT_VIEWS_JTE)
+                .compile());
+        if (generatorContext.getLanguage() == Language.KOTLIN) {
```
dstepanov, line 88: Why are both dependencies with a scope compile? I would expect the view dependency as `implementation` and the other as `compileOnly` - in Gradle terminology

sdelamo: I changed it to compileOnly in the module https://github.com/micronaut-projects/micronaut-views/pull/877

dstepanov: Yes, but my understanding is that the dependency is only needed as a compile plugin so in starter it should be added as compile only too.
Are you sure we need to have that dependency in Micronaut Views?

sdelamo: > Are you sure we need to have that dependency in Micronaut Views?

What dependency?

dstepanov: You wrote that you changed it to `compileOnly`

dstepanov: Regarding this PR, it's a bit strange to have micronaut-views and kotlin with the same scope

PR #2618, opened 2024-09-20
Switch to new Gradle Shadow plugin

timyates:
Previously this plugin was developed by johnrengelman and published under the ID com.github.johnrengelman.shadow before maintenance was transferred to the GradleUp organization to ensure future development

Functionally it's identical, but it doesn't have deprecation warnings like the old one started having

https://github.com/GradleUp/shadow
https://plugins.gradle.org/plugin/com.gradleup.shadow
https://gradleup.com/shadow/

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -13,7 +13,7 @@ dependencies {
     implementation "com.fasterxml.jackson.core:jackson-databind:2.17.1"
     implementation "io.sdkman:gradle-sdkvendor-plugin:3.0.0"
     implementation "org.asciidoctor:asciidoctor-gradle-jvm:2.4.0"
-    implementation "com.github.johnrengelman:shadow:8.1.1"
+    implementation "com.gradleup.shadow:shadow-gradle-plugin:8.3.2"
     implementation "com.microsoft.azure:azure-functions-gradle-plugin:1.16.1"
 }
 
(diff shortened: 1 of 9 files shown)


timyates:
Seems to be failing on a TestContainers/SQLServer issue which I don't believe is of my doing 😉

radovanradic:
> Seems to be failing on a TestContainers/SQLServer issue which I don't believe is of my doing 😉

Yes, it started failing for other projects as well. For sql and data we used different image `image-name: mcr.microsoft.com/mssql/server:2022-latest` and it worked.

timyates:
I guess that change will need to go into https://github.com/micronaut-projects/micronaut-test-resources/blob/2.6.x/test-resources-jdbc/test-resources-jdbc-mssql/src/main/java/io/micronaut/testresources/mssql/MSSQLTestResourceProvider.java

radovanradic:
Yes, unless there is a way to override (temporary solution) in application-test.yml or via `@Property` until test-resources is changed.

melix (commented):
Sounds like a reasonable change, however I think we should probably go into the Gradle plugin and make sure we change it there too, at least in tests.

scprek:
I made an issue in the gradle plugin, but then found this.. I'll leave open? https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/1080

timyates:
I wonder if this should be considered now https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/1081 is merged into the gradle plugin 🤔

melix:
Still needs a release

timyates:
True.... how soon I forget 😉

PR #2662, opened 2024-11-21
give a new look to openapi

CtrlAltFly:
added ui theme as material


sdelamo:
> added ui theme as material

@shekarroy15 is there any problem with the previous default cc @altro3

altro3:
don't think there is problem. @shekarroy15  just wanted to change the default theme :-). Right?

CtrlAltFly:
oh absolutely.. no problems seen. 
in a project i used this theme and it looks much cooler. so why not have it as default for folks who are going to use it hereafter was the thought. 
@sdelamo @altro3

altro3:
In general, I don't see any problems with this. And some themes are actually made better than the default one.

PR #2683, opened 2025-01-02
For Maven, if a dependency with runtime and test, a compile dependency is added

sdelamo:
if a dependency is added to the scopes runtime and test in maven. The dependency should be in compile only for maven.

We should not generate this: 

```xml
<dependency>
  <groupId>ch.qos.logback</groupId>
  <artifactId>logback-classic</artifactId>
  <scope>runtime</scope>
</dependency>
<dependency>
  <groupId>ch.qos.logback</groupId>
  <artifactId>logback-classic</artifactId>
  <scope>test</scope>
</dependency>
```

But this: 

```xml
<dependency>
  <groupId>ch.qos.logback</groupId>
  <artifactId>logback-classic</artifactId>
  <scope>compile</scope>
</dependency>
```

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/build/dependencies/Dependency.java
+++ b/starter-core/src/main/java/io/micronaut/starter/build/dependencies/Dependency.java
@@ -210,6 +210,20 @@ public final class Dependency implements Coordinate {
         return annotationProcessorPriority;
     }
 
+    public static Dependency of(Dependency dep, Scope scope) {
+        return new Dependency(scope,
+                dep.getGroupId(),
+                dep.getArtifactId(),
+                dep.getVersion(),
+                dep.getVersionProperty(),
+                dep.requiresLookup(),
+                dep.isAnnotationProcessorPriority(),
+                dep.getOrder(),
+                dep.isPom(),
+                dep.getExclusions(),
+                dep.getSubstitutions());
+    }
+
     public static class Builder {
 
         private Scope scope;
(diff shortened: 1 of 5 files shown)


melix:
I don't understand this fix. A `compile` dependency will make the dependency part of the transitive compile dependencies, which will leak to consumers. Having both `runtime` and `test` dependencies is more accurate, since these are really runtime only deps.

alvarosanchez:
Agree with @melix. Moreover, the extra `test` scope is redundant since `runtime` dependencies are already in the `test` classpath

sdelamo:
In maven, what we are seeing is that if you have: 

```xml
<dependency>
  <groupId>ch.qos.logback</groupId>
  <artifactId>logback-classic</artifactId>
  <scope>runtime</scope>
</dependency>
<dependency>
  <groupId>ch.qos.logback</groupId>
  <artifactId>logback-classic</artifactId>
  <scope>test</scope>
</dependency>
```

then there is no logback at runtime. it seems the dependency is set as test scope only. 
You can see it in action in this guide: https://guides.micronaut.io/latest/micronaut-server-filter-request-maven-java.html

alvarosanchez:
Probably the latter overrides the former. The extra `test` scope needs to be removed

sdelamo:
@alvarosanchez and @melix 

I have changed the PR. 

## For Maven 

These dependencies: 

```xml
<dependency>
  <groupId>ch.qos.logback</groupId>
  <artifactId>logback-classic</artifactId>
  <scope>runtime</scope>
</dependency>
<dependency>
  <groupId>ch.qos.logback</groupId>
  <artifactId>logback-classic</artifactId>
  <scope>test</scope>
</dependency>
```

get converted to:

```xml
<dependency>
  <groupId>ch.qos.logback</groupId>
  <artifactId>logback-classic</artifactId>
  <scope>runtime</scope>
</dependency>
```

## For Gradle 

These dependencies: 

```xml
runtimeOnly("ch.qos.logback:logback-classic")
testImplementation("ch.qos.logback:logback-classic")
```

it remains the same:

```xml
runtimeOnly("ch.qos.logback:logback-classic")
testImplementation("ch.qos.logback:logback-classic")
```

alvarosanchez (approved):

melix (approved):

PR #2703, opened 2025-02-14
Features backed by Open rewrite recipes

sdelamo:


sdelamo:
@alvarosanchez this PR has a lot of package imports changes because I extracted a micronaut-stater-sdk module. Please, ignore those for now. 

I created a proof of concept of a starter feature consuming a [OpenRewrite declarative recipe](https://docs.openrewrite.org/running-recipes/popular-recipe-guides/authoring-declarative-yaml-recipes),  in particular [Gradle](https://docs.openrewrite.org/recipes/gradle/adddependency) and [Maven dependency recipes](https://docs.openrewrite.org/recipes/maven/adddependency). See the `RecipeDependencyFetcher` api and its usage in the `Mockito` feature. 

What do you think? It is still far from complete but this will allow us to define every feature dependencies with open rewrite recipes. We can then move to configuration recipes etc.

alvarosanchez:
Note that there is [`org.openrewrite.java.dependencies.AddDependency`](https://docs.openrewrite.org/recipes/java/dependencies/adddependency) which works for both Maven and Gradle.

Other than that, as far as I can see (the browser crashes with this gigantic PR), you're not really invoking OpenRewrite, but rather using its declarative recipe format as an alternative, internal way to define dependencies. Which, if I'm not mistaken, would end up going the same route of rendering Rocker templates.

I don't think we should do this. What I meant was to use OpenRewrite to actually perform changes to an existing project, via an additional command (eg `mn add-feature`). With this idea, features may (or may not) have, **in addition** to what they do today, a corresponding OpenRewrite recipe, that would only be used by such command. So that I can run `mn add-feature mockito` in my existing app. With this approach, we can gradually support more and more features in the `add-feature` command, without having to make deep changes in the existing Starter functionality

sdelamo:
@alvarosanchez I think we are on the same page. I  pushed an [extra commit](https://github.com/micronaut-projects/micronaut-starter/pull/2703/commits/362de03998124d3843451857e8803fe08139a5ea) which hopefully clarifies the direction I am suggesting. 

As you said, I want to support both use cases: 
- Project Creation (create, create-app, ... commands),
- Project Modification (add-feature command).

Project Modification will invoke OpenRewrite recipes in an existing project. However, I have not implemented that yet. I have pushed a commit with an `AddFeatureCommand` class with a `TODO` where we will invoke OpenRewrite to execute the rules associated with features. 

To gradually add features I have created a marker interface ` OpenRewriteFeature`. Beans of type ` OpenRewriteFeature` have OpenRewrite rules associated with them. 

For me, the key is to define things only in a single place. That it is the reason why I created the `RecipeDependencyFetcher` API. To support defining the dependencies as open rewrite rules but transform those rules to `Dependency`ies to support the project creation use case.  We will define the dependencies a feature adds on a single place - the OpenRewrite recipe associated with the feature. 

NOTE: Currently, [OpenRewrite does not support adding a gradle dependency to a Gradle file using the Kotlin DSL only the Groovy DSL](https://github.com/openrewrite/rewrite/issues/3291). 

Let me know if you think I am going in the right direction.

starter-openrewrite-recipes/src/main/resources/META-INF/rewrite/rewrite.yml
```diff
@@ -1,13 +1,10 @@
 ---
 type: specs.openrewrite.org/v1beta/recipe
-name: micronaut.starter.feature.mockito.AddDependencyMockito
+name: micronaut.starter.feature.Mockito
```
alvarosanchez: Convention seems to be using package names.

Also, I'd have a YML per main recipe, instead of having a huge `rewrite.yml`. OpenRewrite will detect them all within `META-INF/rewrite`.

starter-core/src/main/java/io/micronaut/starter/openrewrite/OpenRewriteFeature.java
```diff
@@ -0,0 +1,29 @@
...
+import io.micronaut.starter.feature.Feature;
+
+/**
+ * A feature backed by an OpenRewrite recipe.
+ * @author Sergio del Amo
+ */
+public interface OpenRewriteFeature extends Feature {
+
+    @NonNull
+    String getRecipeName();
```
alvarosanchez: I suppose this also supports using [Refaster and imperative recipes](https://docs.openrewrite.org/authoring-recipes/types-of-recipes)?

alvarosanchez:
@sdelamo amazing, then we are in agreement!

starter-openrewrite-recipes/src/main/resources/META-INF/rewrite/rewrite.yml
sdelamo: not sure about using multiple files, we can decided about this later. micronaut resource loader api is not super flexible.

starter-core/src/main/java/io/micronaut/starter/openrewrite/OpenRewriteFeature.java
sdelamo: Yes, I have tried only with declarative recipes. But ideally, we should support every type of recipe.

sdelamo: maybe we could use a default convention such as io.micronaut.starter.feature.${featureName} to ensure unique name

sdelamo:
@alvarosanchez I have pushed a commit of an example of a feature contributing configuration. https://github.com/micronaut-projects/micronaut-starter/pull/2703/commits/d71a1e77ab4da85a2cf74a78830d89f18ff70a6e

We are going to need to create our own AddProperty implementation similar to [AddSpringProperty](https://github.com/openrewrite/rewrite-spring/blob/main/src/main/java/org/openrewrite/java/spring/AddSpringProperty.java), or [AddQuarkusProperty](https://github.com/openrewrite/rewrite-quarkus/blob/main/src/main/java/org/openrewrite/quarkus/AddQuarkusProperty.java) as the `AddProperty` I used in my commit [does not allow to provide a `fileMatcher`](https://github.com/openrewrite/rewrite/issues/3394) and in a Micronaut Application the property entry will be added to gradle.properties instead of application.properties in Micronaut gradle app.

PR #2707, opened 2025-03-07
Adding json-schema-generator feature

elifKurtay:
Adds json-schema-generator feature.


starter-core/src/main/java/io/micronaut/starter/feature/json/JsonSchemaGeneratorFeature.java
```diff
@@ -0,0 +1,78 @@
...
+@Requires(property = "micronaut.starter.feature.json.schema.generator.enabled", value = StringUtils.TRUE, defaultValue = StringUtils.TRUE)
+@Singleton
+public class JsonSchemaGeneratorFeature implements Feature {
+
+    public static final String NAME = "json-schema-generator";
+
+    private static final Dependency JSON_SCHEMA_GENERATOR_DEPENDENCY = MicronautDependencyUtils.jsonSchemaDependency()
+            .artifactId("micronaut-json-schema-generator")
+            .compile()
+            .build();
```
andriy-dmytruk: I don't think this should be added as a compile dependency. 

I think we should add the plugins, add a simple example schema file and a sample schema configuration for the plugins (similarly to how it is done in the guide).

starter-core/src/main/java/io/micronaut/starter/feature/json/JsonSchemaGeneratorFeature.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public String getTitle() {
+        return "JSON Schema Generator";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Adds Source Generator from JSON Schema to a Micronaut Application";
```
andriy-dmytruk: Could this be "Adds capability to generate source classes from JSON schema definitions"?

Also, could you please change the description of the `JsonSchemaFeature`, so the difference is clear.

andriy-dmytruk (approved):

PR #2762, opened 2025-05-29
fix azure functions dependency scope

ayoubouria:
- Set the `azure-functions-java-library` Maven dependency scope to `compile` so it’s packaged into the build and available at runtime, preventing `NoClassDefFoundError`.

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/function/azure/AbstractAzureFunction.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/function/azure/AbstractAzureFunction.java
@@ -55,6 +55,13 @@ public abstract class AbstractAzureFunction extends AbstractFunctionFeature impl
 
     public static final String GROUP_ID_COM_MICROSOFT_AZURE_FUNCTIONS = "com.microsoft.azure.functions";
     public static final String ARTIFACT_ID_AZURE_FUNCTIONS_JAVA_LIBRARY = "azure-functions-java-library";
+    public static final String AZURE_FUNCTIONS_EXTENSION_VERSION = "~4";
+
+    public static final Dependency DEPENDENCY_AZURE_FUNCTIONS_JAVA_LIBRARY = Dependency.builder()
+            .groupId(GROUP_ID_COM_MICROSOFT_AZURE_FUNCTIONS)
+            .artifactId(ARTIFACT_ID_AZURE_FUNCTIONS_JAVA_LIBRARY)
+            .compile()
+            .build();
 
     public static final String NAME = "azure-function";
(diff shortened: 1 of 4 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/function/azure/AbstractAzureFunction.java
```diff
@@ -55,6 +55,13 @@ public abstract class AbstractAzureFunction extends AbstractFunctionFeature impl
 
     public static final String GROUP_ID_COM_MICROSOFT_AZURE_FUNCTIONS = "com.microsoft.azure.functions";
     public static final String ARTIFACT_ID_AZURE_FUNCTIONS_JAVA_LIBRARY = "azure-functions-java-library";
+    public static final String AZURE_FUNCTIONS_EXTENSION_VERSION = "~4";
```
sdelamo, line 58: Use an exact version and put it in the toml file.

sdelamo (changes requested):

starter-core/src/main/java/io/micronaut/starter/feature/function/azure/AbstractAzureFunction.java
ayoubouria: Azure docs recommends setting `FUNCTIONS_EXTENSION_VERSION` (which specifies the Azure Functions runtime) to the major version, so that Azure automatically installs all minor and patch updates

sdelamo: Thanks, I understand. However, I don't want minor or patch updates automatically installed. I want an exact version for reliable builds.

ayoubouria: Choosing the host exact version is managed by Azure and manage it manually can be difficult:  

- when creating an Azure function using CLI, portal or API, you cannot specify the exact version the only accepted value is '~4'  and then Azure chooses the host version based on Java version/OS/hosting plan, so the version can vary based on them
- also Azure allows pinning the specific version after creating the function (except for consumption plan linux function, it doesn't support pinning a specific version It always auto-updates),  and pinning to a specific version differs between Windows and Linux
- minor versions are [periodically removed from Functions](https://github.com/Azure/app-service-announcements/issues) and they are short lived, if a minor version is removed Azure automatically change to the closest available minor version so Azure will change the version eventually

So if we used an exact version in the template, the user who wants to deploy a Micronaut app as an Azure function needs to check if that version is not removed, check if it’s compatible with the Java version and OS, then pin the version in the function and restart it, and only then deploy

PR #2768, opened 2025-06-20
Revert "ignore SQLServer test"

sdelamo:
This reverts commit 07920da7bf494fd94c8bbab2eace19b1c5fd0d7e.


sdelamo:
SQLServer are disabled. To enable them, we could set a system property. 

```
        // In netty 4.2.2 SslContextBuilder defaults endpointIdentificationAlgorithm to "HTTPS" if system property
        // "io.netty.handler.ssl.defaultEndpointVerificationAlgorithm" is not set. This makes sun.security.util.HostnameChecker.match(String expectedName, X509Certificate cert, boolean chainsToPublicCA)
        // method to throw error "javax.net.ssl.SSLHandshakeException: No name matching localhost found" and this is workaround.
        // Which could be documented or suggested if users experience this issue
System.setProperty("io.netty.handler.ssl.defaultEndpointVerificationAlgorithm", "NONE")
```

PR #2771, opened 2025-06-21
add common compress for testcontainers

sdelamo:
see:
https://github.com/testcontainers/testcontainers-java/issues/8798 https://github.com/testcontainers/testcontainers-java/issues/8338#issuecomment-1992649517


sdelamo:
@graemerocher @melix @radovanradic we have to keep an eye on this to see if this is required also in Test Resources or downstream.

PR #2854, opened 2025-09-29
add java 25 as an option

### LINKED ISSUE #2853: Add Java 25 as an option
Image

lissam1:


starter-core/src/main/java/io/micronaut/starter/feature/build/gradle/Gradle.java
```diff
@@ -36,26 +36,48 @@
...
     public static final boolean DEFAULT_USER_VERSION_CATALOGUE = false;
     protected static final GradlePlugin GROOVY_GRADLE_PLUGIN = GradlePlugin.builder().id("groovy").build();
 
-    protected static final String WRAPPER_JAR = "gradle/wrapper/gradle-wrapper.jar";
-    protected static final String WRAPPER_PROPS = "gradle/wrapper/gradle-wrapper.properties";
+    private static final String GRADLE = "gradle";
+    private static final String DASH = "-";
+    private static final String SLASH = "/";
+    private static final String GRADLE_8_14_2 = "8.14.2";
+    private static final String GRADLE_9_1_0 = "9.1.0";
```
melix, line 59: Would be easier if these versions were not managed in sources. Similarly, the presence of the Gradle wrapper in resources is a bit silly. Maybe not for this PR, but I think we should refactor this so that the versions are handled in the build, and that the build is responsible for _generated resources directories_ which contain the wrapper.

melix (approved):
Did you actually check that the resulting generated projects work? It's a difficult problem since running on 9.1.0 may itself not work correctly with the build plugins...

sdelamo:
> Did you actually check that the resulting generated projects work? It's a difficult problem since running on 9.1.0 may itself not work correctly with the build plugins...

I have not done exhaustive testing, but this works: 

```
./gradlew micronaut-cli:run --args="create-app --jdk 25 -b gradle temp4"
cd starter-cli/temp4 
sdk use java 25-graal;export GRAALVM\_HOME=/Users/sdelamo/.sdkman/candidates/java/25-graal
./gradlew build
```

What I did for maven: 

```
./gradlew micronaut-cli:run --args="create-app --jdk 25 -b maven temp3”
cd starter-cli/temp3 
sdk use java 25-graal;export GRAALVM\\\_HOME=/Users/sdelamo/.sdkman/candidates/java/25-graal
./mvnw verify
```

alvarosanchez (approved):

PR #2850, opened 2025-09-22
Feature mn update

lissam1:

PR #2857, opened 2025-10-03
Fix gradle runner: feature mn update

lissam1:
Replaced Gradle Tooling API usage (GradleConnector/ProjectConnection) with external Gradle execution via ProcessBuilder for native-image compatibility.

### DIFF
--- a/starter-cli/src/main/java/io/micronaut/starter/cli/util/openrewrite/gradle/GradleOpenRewriteRecipesRunner.java
+++ b/starter-cli/src/main/java/io/micronaut/starter/cli/util/openrewrite/gradle/GradleOpenRewriteRecipesRunner.java
@@ -30,9 +30,7 @@ import java.util.ArrayList;
 import java.io.InputStream;
 
 import java.io.File;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.function.Consumer;
 
 @Named("gradle")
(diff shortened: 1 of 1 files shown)


starter-cli/src/main/java/io/micronaut/starter/cli/util/openrewrite/gradle/GradleOpenRewriteRecipesRunner.java
```diff
@@ -59,22 +57,53 @@ public void run(@NonNull List<String> recipes,
...
+            if (wrapper.isFile()) {
+                if (isWindows) {
+                    cmd.add("cmd");
+                    cmd.add("/c");
+                    cmd.add(wrapper.getAbsolutePath());
+                } else {
+                    wrapper.setExecutable(true);
+                    cmd.add(wrapper.getAbsolutePath());
+                }
+            } else {
```
alvarosanchez, line 73: This assumes they have Gradle installed, which is a big assumption. Perhaps we should just restrict this to projects where the wrapper is installed, and if it's not, ask them to install it.

@melix wdyt?

lissam1: context : https://github.com/micronaut-projects/micronaut-starter/pull/2850#issuecomment-3612145247

melix: Right. So we had a pairing session with Issam yesterday and I confirm that the tooling API will not work when bundled in a native image, because of the use of dynamic classloading.

I agree with @alvarosanchez that we should check if the wrapper is available, and from what I'm seeing, it is already the case. So it's just a matter of issuing a warning if not.

starter-cli/src/main/java/io/micronaut/starter/cli/util/openrewrite/gradle/GradleOpenRewriteRecipesRunner.java
```diff
@@ -89,16 +118,15 @@ public void run(@NonNull List<String> recipes,
...
-        Map<String, String> systemProperties = new HashMap<>();
-        Map<String, Object> configurationSystemProperties = configuration.getSystemProperties();
-        for (String k : configurationSystemProperties.keySet()) {
-            Object value = configurationSystemProperties.get(k);
-            if (value != null) {
-                systemProperties.put(k, value.toString());
-            }
 
+    private void pump(InputStream in, Consumer<String> consumer) {
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8))) {
```
melix, line 123: This needs to be tested, I think that under Windows this can lead to eternal waiting because no new line will be returned, depending on the buffer size.

PR #2867, opened 2025-10-09
Mcp studio feature

### LINKED ISSUE #2864: Add MCP Features
We need to add several features to starter for the Micronaut MCP module. 

https://micronaut-projects.github.io/micronaut-mcp/snapshot/guide/

# `mcp-stdio` feature

We need a feature named `mpc-stdio`. 

- It should add the dependency with `compile` scope `io.micronaut.mcp:micronaut-mcp-server-java-sdk`
- It should add the configuration `micronaut.mcp.server.info.name` with value `mcpdemo`
- It should add the configuration `micronaut.mcp.server.info.version` with value `0.0.1`
- It should add the configuration `micronaut.mcp.server.transport` with value `STDIO`. It should be `uppercase`
- [It should configure logback.xml to log to `std.err`](https://micronaut-projects.github.io/micronaut-mcp/snapshot/guide/#stdio). 
- It should disable the Micronaut Banner
- It should remove the `runtime("netty")` when using gradle. And it should not add the `micronaut-http-server-netty` when using maven. 
  
# `mcp-http` feature

We need a feature named `mpc-http`. 

- It should add the dependency : `io.micronaut.mcp:micronaut-mcp-server-java-sdk`
- It should add the configuration `micronaut.mcp.server.info.name` with value `mcpdemo`
- It should add the configuration `micronaut.mcp.server.info.version` with value `0.0.1`
- It should add the configuration `micronaut.mcp.server.transport` with value `HTTP`. It should be `uppercase`

# `mcp-client-java-sdk` feature

We need a feature named `micronaut-mcp-client-java-sdk`. 

- It should add the dependency with `test` scope `io.micronaut.mcp:micronaut-mcp-client-java-sdk`

# `mcp-client-langchain4j` feature

We need a feature named `micronaut-mcp-client-langchain4j`. 

- It should add the dependency with `test` scope `io.micronaut.mcp:micronaut-mcp-client-langchain4j`

All the features should be tested with tests in `starter-core`

lissam1:


starter-core/src/main/java/io/micronaut/starter/feature/mcp/McpStdio.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    public boolean supports(ApplicationType applicationType) {
+        return true;
+    }
+
+    @Override
+    public void processSelectedFeatures(FeatureContext featureContext) {
+            featureContext.exclude(Netty.class::isInstance);
+    }
```
lissam1, line 62: > It should remove the runtime("netty") when using gradle. And it should not add the micronaut-http-server-netty when using maven

i think this will work for both maven and gradle

sdelamo: I think it is more correct https://github.com/micronaut-projects/micronaut-starter/pull/2867/commits/2569c6bf4cf0d0b45df8ff5c8d6b09b978f7a9f4

PR #2886, opened 2025-10-20
feature: micronaut test netty leak

sdelamo:
Adds feature for https://github.com/micronaut-projects/micronaut-test/pull/1291

### DIFF
--- a/starter-core/src/main/java/io/micronaut/starter/feature/test/MicronautTestNettyLeak.java
+++ b/starter-core/src/main/java/io/micronaut/starter/feature/test/MicronautTestNettyLeak.java
@@ -30,11 +30,12 @@ import io.micronaut.starter.options.Options;
 import io.micronaut.starter.options.TestFramework;
 import jakarta.inject.Singleton;
 
+import java.util.Map;
 import java.util.Set;
 
 @Requires(property = "micronaut.starter.feature.micronaut-test-netty-leak.enabled", value = StringUtils.TRUE, defaultValue = StringUtils.TRUE)
 @Singleton
-public class MicronautTestNettyLeak implements DefaultFeature {
+public class MicronautTestNettyLeak implements DefaultFeature, JunitPlatformPropertyProvider {
     private static final Dependency DEPENDENCY_MICRONAUT_TEST_NETTY_LEAK = MicronautDependencyUtils.testDependency()
             .artifactId("micronaut-test-netty-leak")
             .testRuntime()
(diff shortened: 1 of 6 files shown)


starter-core/src/main/java/io/micronaut/starter/feature/test/MicronautTestNettyLeak.java
```diff
@@ -0,0 +1,92 @@
...
+import io.micronaut.starter.feature.Category;
+import io.micronaut.starter.feature.DefaultFeature;
+import io.micronaut.starter.feature.Feature;
+import io.micronaut.starter.options.Options;
+import io.micronaut.starter.options.TestFramework;
+import jakarta.inject.Singleton;
+
+import java.util.Set;
+
+@Requires(property = "micronaut.starter.feature.micronaut-test-netty-leak.enabled", value = StringUtils.TRUE, defaultValue = StringUtils.TRUE)
```
yawkat, line 32: i dont think this is useful enough for most people to become a default

sdelamo: we add this `@Requires` for every feature.  It enables users to disable the feature.

yawkat:
I think we also need some code to enable extension detection for junit 5?

sdelamo:
> I think we also need some code to enable extension detection for junit 5?

I created a `src/test/resources/junit-platform.properties` with  `junit.jupiter.extensions.autodetection.enabled=true`

https://docs.junit.org/current/user-guide/#extensions-registration-automatic-enabling

starter-core/src/main/java/io/micronaut/starter/feature/test/MicronautTestNettyLeak.java
```diff
@@ -0,0 +1,101 @@
...
+
+    protected void addDependency(GeneratorContext generatorContext) {
+        if (generatorContext.hasFeature(Netty.class)) {
+            generatorContext.addDependency(DEPENDENCY_MICRONAUT_TEST_NETTY_LEAK);
+        }
+    }
+
+    @Override
+    public boolean shouldApply(ApplicationType applicationType, Options options, Set<Feature> selectedFeatures) {
+        return options.getTestFramework() == TestFramework.JUNIT || options.getTestFramework() == TestFramework.SPOCK;
```
yawkat: it should not be on by default for these frameworks

sdelamo: why not?

yawkat: leak detection is too strict and not useful for most people

sdelamo: but it is strict, in the sense of, of given false positives?

yawkat: yes

yawkat (approved):

yawkat:
thanks!

PR #2872, opened 2025-10-10
Migrating to the Develocity plugin

lissam1:

PR #2929, opened 2025-11-06
Update Develocity Feature description

guylabs:
Add a more suitable description of Develocity that better showcases its value.

Follow-up on #2872.


guylabs:
Hi @lissam1 @sdelamo, would it be possible to update the description of Develocity? That would be great! Many thanks!

sdelamo (approved):

sdelamo:
thanks for the contribution @guylabs

